PR #3, opened 2018-03-08
LA-326 Testext environment support; expanded Redshift code and tests

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-326

Since these external tests require access to private configs (and may be slow), they don't run under Travis. They could conceivably be incorporated into a CodePipeline run, as CalCentral/Junction does on Bamboo.


johncrossman:
image

PR #5, opened 2018-03-21
LA-320 Canvas Data API client, test coverage

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-320

`nessie.lib.mockingbird` and `nessie.lib.util` are straight copy-paste from their BOAC equivalents. At some point, if it looks like we're going to be keeping around a stable of Python projects, we might look at extracting shared dependencies into their own repo.


johncrossman:
image

PR #6, opened 2018-03-21
LA-338 Master-to-worker job dispatch

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-338

Library code supporting non-GET HTTP methods can be copied back to BOAC.


johncrossman:
She _is_ a hard-working Nessie.

PR #7, opened 2018-03-22
LA-340 S3 cleanup routine for obsolete files

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-340


johncrossman:
Squeaky clean.

PR #9, opened 2018-03-23
LA-322 End-to-end logic for Canvas snapshot sync

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-322

The very large fixtures/canvas_data_file_sync.json is slimmed down somewhat for faster test runs.


sandeepmjay:
LGTM @paulkerschen. Shipping it

PR #10, opened 2018-04-11
LA-321 Add creation of Canvas data schema in Redshift

raydavis:
https://jira.ets.berkeley.edu/jira/browse/LA-321

No tests yet, but there are enough oddities here to call for an early review.


johncrossman:
Nice work, @raydavis

PR #11, opened 2018-04-11
 LA-321 Add BOAC analytics schema creation

raydavis:
https://jira.ets.berkeley.edu/jira/browse/LA-321


johncrossman:
All looks good, @raydavis  Travis would like to have a word...

PR #12, opened 2018-04-12
LA-368 Bring BOAC and Canvas schema templates up to date with Data Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/LA-368


johncrossman:
That's some prize-winning SQL!  Thanks @raydavis

PR #13, opened 2018-04-12
LA-323 Determine the current date string based on LA timezone

raydavis:
The easiest bit of https://jira.ets.berkeley.edu/jira/browse/LA-323


paulkerschen:
It's five o'clock somewhere.

PR #21, opened 2018-04-20
LA-400 Improve performance of assignment timeliness query

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-400


johncrossman:
You had me at "Improve performance". Thanks @paulkerschen

PR #22, opened 2018-04-20
LA-397 Make SIS enrollment status available to loch queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-397

Three commits because the work comes in three parts:

1. Create an external schema in Redshift for our SIS data snapshots;
2. Generate some persistent internal tables to bring together Canvas and SIS data. (I too am suspicious of creating schemata all over the place, but we did seem to need at least one more for generally applicable internal tables that have nothing to do with BOAC as such.)
3. Take some load off BOAC-specific table creation by allowing it to pull data from the intermediate tables, including SIS enrollment status. Queries from the BOAC side can then filter their result sets by enrollment status if desired.

### DIFF
--- a/config/default.py
+++ b/config/default.py
@@ -54,9 +54,11 @@ CANVAS_DATA_HOST = 'foo.instructure.com'
 LOCH_S3_BUCKET = 'bucket_name'
 LOCH_S3_REGION = 'us-west-2'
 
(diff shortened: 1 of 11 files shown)


raydavis:
This makes the Loch seem a lot less mirage-like and Nessie seem a lot less mythical. Thank you, @paulkerschen

PR #23, opened 2018-04-23
LA-401, default listener (port 80) is enabled, but will redirect to 443 per Apache config

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/LA-401

As seen in BOAC: https://github.com/ets-berkeley-edu/boac/blob/master/.ebextensions/00_ami.config


paulkerschen:
Ah.

PR #24, opened 2018-04-23
LA-402, route '/' will go to '/api/ping'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/LA-402


paulkerschen:
Twice the communication.

PR #25, opened 2018-04-24
LA-407, new SG settings for custom VPC

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/LA-407

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -33,3 +33,4 @@ artifacts:
   - 'nessie/**/*'
   - 'requirements.txt'
   - 'run.py'
+  - 'scripts/**/*'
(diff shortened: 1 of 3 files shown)


buildspec.yml
```diff
@@ -33,3 +33,4 @@ artifacts:
   - 'nessie/**/*'
   - 'requirements.txt'
   - 'run.py'
+  - 'scripts/**/*'
```
johncrossman: `scripts/**/*` is not currently required but, eventually, we might want it bundled with deployments.

sandeepmjay:
Looks good. Merging @johncrossman

PR #26, opened 2018-04-25
LA-318 Support job scheduling from config files

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-318

Note that configs defined here will constitute our first split between master and worker environments: only master should wield the power of JOB_SCHEDULING_ENABLED.


raydavis:
apscheduler does look nice. I wonder whether the scripts should be run as separately jobs or as a series of consecutive steps. But we can always change that later.

PR #28, opened 2018-04-26
LA-411 Update testext suite to use a shared S3 bucket

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-411


raydavis:
Shall I compare thee to a PREFIX_TESTEXT?

PR #30, opened 2018-04-27
LA-413 Suppress PYTHONPATH for AWS CLI command in .ebextensions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-413


raydavis:
I never liked that environment variable anyway.

PR #36, opened 2018-05-03
LA-419 Support queued background job processing

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-419

This seems to work locally; let's try it in The Cloud.


sandeepmjay:
Power of cloud compels you !

PR #39, opened 2018-05-04
NOJIRA Nessie starts out at version 2.0

paulkerschen:
We're ahead out of the gate!

### DIFF
--- a/nessie/__init__.py
+++ b/nessie/__init__.py
@@ -24,4 +24,4 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """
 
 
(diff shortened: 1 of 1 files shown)


johncrossman:
Smart @paulkerschen!  You're thinking like Larry Ellison: "The very first version was Oracle version 2, we knew no one would want to buy version 1."  But Travis ain't convinced.

paulkerschen:
@johncrossman - https://jira.ets.berkeley.edu/jira/browse/LA-443 created for the annoying intermitted failure.

PR #40, opened 2018-05-08
LA-443 Plug intermittent test failure with fixture

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-443

Moto, our mock AWS library, prefers HTTP mocks to come by way of `responses` rather than `httpretty`. It would be nice to get all our HTTP mocks consistent, but this stops the intermittent failure for now.


johncrossman:
I send this on, and I joy.

PR #41, opened 2018-05-09
LA-428 Verify external schemas after drop and recreation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-428

If anything is amiss, returning `False` from these jobs will abort subsequent jobs and leave BOAC's materialized views intact.


raydavis:
Simple and important -- thank you, @paulkerschen !

PR #45, opened 2018-05-17
NS-46 Include SIS course title in Nessie

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-46

Next down in the waterfall is BOAC itself.


raydavis:
Our unit tests don't have sharp teeth when it comes to schema creation, but my test run of the change produced the expected results.

sandeepmjay:
Shipping it @raydavis !

PR #48, opened 2018-05-22
NS-46 Improve access to SIS section and enrollment data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-46

Another shuffle of the external-data deck.

### DIFF
--- a/nessie/sql_templates/create_intermediate_schema.template.sql
+++ b/nessie/sql_templates/create_intermediate_schema.template.sql
@@ -25,12 +25,34 @@
 
 CREATE SCHEMA IF NOT EXISTS {redshift_schema_intermediate};
 
-DROP TABLE IF EXISTS {redshift_schema_intermediate}.course_sections;
+/*
+ * Copy S3 external data to Redshift table for faster client queries.
+ */
+
+DROP TABLE IF EXISTS {redshift_schema_intermediate}.sis_enrollments;
+
+CREATE TABLE {redshift_schema_intermediate}.sis_enrollments
+INTERLEAVED SORTKEY (sis_term_id, sis_section_id, ldap_uid)
+AS (
+    SELECT
+        en.term_id AS sis_term_id,
+        en.section_id AS sis_section_id,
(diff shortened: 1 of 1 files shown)


raydavis:
I'm closing this to check on a possibly missing piece...

raydavis:
Phew. False alarm, due to the SIS API's many flavors of "displayName".

nessie/sql_templates/create_intermediate_schema.template.sql
```diff
@@ -65,26 +111,20 @@ AS (
         s.name AS canvas_section_name,
         et.name AS canvas_course_term,
         extracted_section_ids.sis_term_id,
-        extracted_section_ids.sis_section_id,
-        sc.course_display_name AS sis_course_name,
-        sc.course_title AS sis_course_title,
-        sc.instruction_format AS sis_instruction_format,
-        sc.section_num AS sis_section_num,
-        sc.is_primary AS sis_primary
+        extracted_section_ids.sis_section_id
```
paulkerschen: @raydavis Sorry for the drive-by - I think the bug you found in NS-46 isn't that Canvas-less SIS sections are missing, but that Canvas-less SIS sections have null `sis_term_id` and `sis_section_id`. Nessie SQL is pulling from the wrong side of the join and should (I think) be using `sc`, not `extracted_section_ids`. This is just a hunch; apologies if you're way ahead of me.

raydavis:
Oh, good eye, @paulkerschen ! (Wish I could say as much for the clavicle.) I'll amend `course_sections` accordingly. But I think I'll leave `sis_enrollments` there rather than trying a similarly complex outer join in `active_student_enrollments` -- I'm in a maze of twisty little BOAC data structures, and scared to start again from the beginning.

raydavis:
PR amended.

raydavis:
Back to the kitchen again. I've been so focused on the analytics side of the house that I lost sight of the UX's dependency on section instructors-and-schedules.

PR #51, opened 2018-05-30
BOAC-908 Parse SIS courses.is_primary as Boolean rather than string

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-908


raydavis:
I confirmed locally that this fixes the BOAC issue.

PR #54, opened 2018-06-05
BOAC-917 A student without scores might lack a course_score_fact row

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-917

When that happens, we still need to return `last_activity_at` and `sis_enrollment_status`.


johncrossman:
Power to the LEFTies.

PR #59, opened 2018-06-12
NS-101, nessie/__init__.py gets the SQLAlchemy basics

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-101


paulkerschen:
Fire burn, and cauldron bubble!

PR #60, opened 2018-06-12
NS-85 Client and tests for SIS student API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-85

Swiped from BOAC all round.

### DIFF
--- a/nessie/externals/sis_student_api.py
+++ b/nessie/externals/sis_student_api.py
@@ -1,0 +1,61 @@
+"""
+Copyright ©2018. The Regents of the University of California (Regents). All Rights Reserved.
+
+Permission to use, copy, modify, and distribute this software and its documentation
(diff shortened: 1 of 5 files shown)


nessie/externals/sis_student_api.py
```diff
@@ -0,0 +1,61 @@
...
+"""Official access to student data."""
+
+
+from flask import current_app as app
+from nessie.lib import http
+from nessie.lib.mockingbird import fixture
+# TODO from nessie.models.json_cache import stow
+
+
+# TODO @stow('sis_student_api_{cs_id}')
```
paulkerschen: `@stow` decorators can be uncommented once we have a Postgres-backed JsonCache in place.

raydavis:
Steal from the best.

PR #63, opened 2018-06-12
NS-86 Client, tests and fixtures for SIS degree progress API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-86

Once again, `@stow` temporarily disabled.


johncrossman:
Progress indeed.

PR #66, opened 2018-06-12
NS-104 Add JsonCache

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-104

We may not need as much staging-table fanciness on a machine that isn't trying to serve BOAC at the same time, but there doesn't seem to be any harm in keeping it around for now.


raydavis:
Play the old songs.

johncrossman:
As they say in Boston, I'm waiting for the T.

PR #68, opened 2018-06-12
NS-84, initial CalNet integration

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-84

I dealt with TODOs in ASC integration.


paulkerschen:
The craziest fixtures of all. Waiting for T-Rex.

PR #69, opened 2018-06-13
NS-108 Remove page views logic from Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-108

It was a sublime hunk of SQL but we'll always have GitHub for the memories...


raydavis:
I'll amend this to "Move page views logic from BOAC schema" and submit a new PR.

PR #74, opened 2018-06-14
NS-102, use 'flask initdb' to create RDS schema

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-102


paulkerschen:
Handy!

PR #77, opened 2018-06-14
NS-89 Job schedule API endpoint

paulkerschen:

It's no `crontab -l`, but I hope it serves. A global scheduler object gives us something to grab when needed, though it probably won't be enough to prevent multiple scheduler instances in different processes. @raydavis's suggestion in https://jira.ets.berkeley.edu/jira/browse/NS-107 of a Postgres advisory lock seems like a good way to handle it.

### DIFF
--- a/nessie/api/job_controller.py
+++ b/nessie/api/job_controller.py
@@ -34,10 +34,28 @@ from nessie.jobs.create_sis_schema import CreateSisSchema
 from nessie.jobs.generate_boac_analytics import GenerateBoacAnalytics
 from nessie.jobs.generate_intermediate_tables import GenerateIntermediateTables
 from nessie.jobs.resync_canvas_snapshots import ResyncCanvasSnapshots
+from nessie.jobs.scheduling import get_scheduler
 from nessie.jobs.sync_canvas_snapshots import SyncCanvasSnapshots
 from nessie.jobs.sync_file_to_s3 import SyncFileToS3
 from nessie.lib.http import tolerant_jsonify
 from nessie.lib.metadata import update_canvas_sync_status
+
+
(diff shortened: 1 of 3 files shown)


nessie/jobs/scheduling.py
```diff
@@ -38,7 +45,8 @@ def initialize_job_schedules(app):
     from nessie.jobs.resync_canvas_snapshots import ResyncCanvasSnapshots
     from nessie.jobs.sync_canvas_snapshots import SyncCanvasSnapshots
 
-    if app.config['JOB_SCHEDULING_ENABLED']:
+    global sched
```
raydavis: Thanks for this -- it makes sense as a first step, anyway.

nessie/api/job_controller.py
```diff
@@ -34,12 +34,30 @@
...
+from nessie.jobs.scheduling import get_scheduler
 from nessie.jobs.sync_canvas_snapshots import SyncCanvasSnapshots
 from nessie.jobs.sync_file_to_s3 import SyncFileToS3
 from nessie.lib.http import tolerant_jsonify
 from nessie.lib.metadata import update_canvas_sync_status
 
 
+@app.route('/api/job/schedule', methods=['GET'])
+def job_schedule():
+    sched = get_scheduler()
```
raydavis: Did you get a chance to try `sched.print_jobs`. Just curious because I haven't tried it myself yet...

paulkerschen: @raydavis I did, and the output is pretty much what you see here, only it doesn't include the job args, which is where our actual task names live.

raydavis: Yeah, probably a good idea to include the task names.

PR #78, opened 2018-06-14
NS-102, shell script does nasty business of running 'flask initdb'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-102

I did necessary `chmod` on the script. Verified in dev.


paulkerschen:
The only nasty thing I like is a nasty groove. Thanks @johncrossman.

PR #80, opened 2018-06-15
NS-94 Safer implementation of DROP EXTERNAL SCHEMA

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-94

Tested against working schemas on dev server.


johncrossman:
Drop that beat, @raydavis

PR #81, opened 2018-06-15
NS-115, give ImportASC the BackgroundJob treatment, with heavy lifting on external/asc

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-115
https://jira.ets.berkeley.edu/jira/browse/NS-83

The ASC `external` is bare bones; the `ImportAscAthletes` job does heavy lifting. For readability, certain util methods live in `lib/asc.py`.


raydavis:
Nicely cut and shuffled, @johncrossman -- thank you.

PR #82, opened 2018-06-15
NS-89 Move scheduler job store to SQLAlchemy; add advisory locks

paulkerschen:

### DIFF
--- a/nessie/jobs/scheduling.py
+++ b/nessie/jobs/scheduling.py
@@ -27,17 +27,27 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """Background job scheduling."""
 
 
+from apscheduler.jobstores.sqlalchemy import SQLAlchemyJobStore
 from apscheduler.schedulers.background import BackgroundScheduler
 
 
+app = None
 sched = None
+
+
+# Postgres advisory locks require numeric ids.
+PG_ADVISORY_LOCK_IDS = {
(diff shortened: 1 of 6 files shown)


nessie/jobs/scheduling.py
```diff
@@ -27,31 +27,45 @@
...
+def initialize_job_schedules(_app):
     from nessie.jobs.create_canvas_schema import CreateCanvasSchema
     from nessie.jobs.create_sis_schema import CreateSisSchema
     from nessie.jobs.generate_boac_analytics import GenerateBoacAnalytics
     from nessie.jobs.generate_intermediate_tables import GenerateIntermediateTables
     from nessie.jobs.resync_canvas_snapshots import ResyncCanvasSnapshots
     from nessie.jobs.sync_canvas_snapshots import SyncCanvasSnapshots
 
+    global app
+    app = _app
```
paulkerschen: `app` has changed from a passed-around argument to another dicey global because with the move to SQLAlchemyJobStore, job arguments must be serializable to binary blobs via `pickle`.

raydavis: Thanks for trying out some alternatives. Maybe we'll be able to get serializable arguments some other way later, but this shouldn't block the merge.

raydavis:
Thanks for getting Postgres advisory locks into the toolbox!

PR #84, opened 2018-06-18
NS-89 POST endpoint to manage job schedules

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-89

Example usage:

`curl -u username:password -d '{"hour": 12, "minute": "30"}' https://app-master-prod.ets-berkeley-nessie.net/api/schedule/job_sync_canvas_snapshots`

To turn the job off:

`curl -u username:password -d '{}' http://app-master-prod.ets-berkeley-nessie.net/api/schedule/job_sync_canvas_snapshots`

PR #85, opened 2018-06-18
NS-89 Preserve modified schedules on app restart

paulkerschen:

Here's an important little follow-up to #84... we don't want to lose our modifications every time the app deploys or restarts.


sandeepmjay:
Looks good @paulkerschen. Merging

PR #86, opened 2018-06-19
NS-120, deploy Nessie configs from S3 to local EC2 instance

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-120

As seen in BOAC.


raydavis:
If it was in BOAC, it's _gotta_ be good!

PR #87, opened 2018-06-19
NS-82, create_asc_schema uses JSON files uploaded to S3 by ASC import job

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-82

* `get_psycopg_cursor` is now a shared db util because _psycopg2_ is necessary to run SQL: `COPY (SELECT ROW_TO_JSON...)` during ASC import.
* `create_asc_schema.template.sql` is unfinished; verifying Redshift schema creation is next step.

### DIFF
--- a/nessie/jobs/import_asc_athletes.py
+++ b/nessie/jobs/import_asc_athletes.py
@@ -23,15 +23,20 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-
+import os
+import tempfile
 from flask import current_app as app
 from nessie import db, std_commit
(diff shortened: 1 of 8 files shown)


nessie/jobs/import_asc_athletes.py
```diff
@@ -90,6 +95,18 @@ def run(self, force=False):
                 else:
                     status['change_counts'].update(_update_from_asc(api_results))
                     asc.confirm_sync(sync_date)
+                    # Dump contents of nessie db (RDS) to JSONs in tmp dir
+                    tmp_dir = f'{tempfile.mkdtemp()}/manifests'
+                    os.mkdir(tmp_dir, 0o777)
```
paulkerschen: 0o, octal literals!

paulkerschen:
psycopg, qu'est-ce que c'est?

PR #92, opened 2018-06-20
NS-92 Log background job status to new metadata table

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-92


johncrossman:
Get a job (and log it). Thx @paulkerschen

PR #93, opened 2018-06-20
NS-125, drop uid and name from students table (RDS) and improve logging

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-125

* UID and name will be pulled from elsewhere.
* Logging `student {index + 1} of {student_count}` is a poor man's progress indicator


paulkerschen:
Horses with no name.

PR #96, opened 2018-06-22
BOAC-982, if ASC API is down then log a proper error message

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-982

"API returned zero results" is wrong if ASC API is down.


nessie/externals/asc_athletes_api.py
```diff
@@ -32,8 +32,9 @@
 def get_asc_feed():
     response = _get_asc_feed_response()
     if not response or not hasattr(response, 'json'):
-        app.logger.error(f'Bad response from ASC Athletes API: {response}')
-        return None
+        error = f'ASC API unexpected response: {response}'
```
raydavis: Thanks, @johncrossman -- this was probably left over from an early `@stow` of the feed.

PR #100, opened 2018-06-27
NS-81 Job to kick off BOAC cache refresh

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-81

If this PR passes muster, there are a couple of operational steps associated with deploy:
- Add BOAC_API_KEY, BOAC_CACHE_REFRESH_URL, and JOB_REFRESH_BOAC_CACHE keys to nessie-master-dev config;
- Remove the corresponding cronjob on junction-dev so that Nessie can take over.

### DIFF
--- a/config/default.py
+++ b/config/default.py
@@ -57,6 +57,9 @@ ASC_ATHLETES_API_URL = 'https://secreturl.berkeley.edu/intensives.php?AcadYr=201
 ASC_ATHLETES_API_KEY = 'secret'
 ASC_THIS_ACAD_YR = '2017-18'
 
+BOAC_API_KEY = 'Regents of the University of California'
+BOAC_CACHE_REFRESH_URL = 'https://ets-boac.example.com/api/refresh_me'
+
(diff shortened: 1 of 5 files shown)


config/default.py
```diff
@@ -57,6 +57,9 @@
 ASC_ATHLETES_API_KEY = 'secret'
 ASC_THIS_ACAD_YR = '2017-18'
 
+BOAC_API_KEY = 'Regents of the University of California'
+BOAC_CACHE_REFRESH_URL = 'https://ets-boac.example.com/api/refresh_me'
```
paulkerschen: The whole URL is stored in configs rather than just the hostname so that it's easy to switch between `/refresh`, `/import_refresh`, etc. as desired.

raydavis: Makes sense. In production, it seems likely that Nessie won't be sending this request until after Nessie has already taken over the import-from-Advisors-Community job, but this lets us test the dev-to-dev connection early.

PR #101, opened 2018-06-27
NS-136, COE data gets Redshift schema with advisor/student mappings

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-136

Notice the TODO: When COE delivers data daily then unleash the daily-hash-in-S3 pattern


paulkerschen:
Look at that Nessie, working ten jobs and staying afloat. Thanks @johncrossman

PR #104, opened 2018-06-28
NS-131 Background job to generate student profile tables in Redshift

paulkerschen:

Let's see how this flies. We loop through all sids in our Calnet stash, create merged profiles and output results to three internal Redshift tables: one with a big JSON payload and two others with extracted queryable values. (I expect that these queries will run somewhat slow compared to Postgres, but our future selves can sort that out.)

Results for our athlete population can be inspected in nessie_dev under the student_pk_test schema.

If this seems workable, the next steps would be to do something similar with student enrollment terms and try to fill out the skimpy test coverage.

### DIFF
--- a/nessie/jobs/generate_merged_profiles.py
+++ b/nessie/jobs/generate_merged_profiles.py
@@ -1,0 +1,150 @@
+"""
+Copyright ©2018. The Regents of the University of California (Regents). All Rights Reserved.
+
+Permission to use, copy, modify, and distribute this software and its documentation
+for educational, research, and not-for-profit purposes, without fee and without a
+signed licensing agreement, is hereby granted, provided that the above copyright
+notice, this paragraph and the following two paragraphs appear in all copies,
+modifications, and distributions.
+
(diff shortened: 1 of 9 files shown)


nessie/jobs/generate_merged_profiles.py
```diff
@@ -0,0 +1,150 @@
...
+from nessie.merged.sis_profile import get_merged_sis_profile
+import psycopg2.sql
+
+
+class GenerateMergedProfiles(BackgroundJob):
+
+    destination_schema = app.config['REDSHIFT_SCHEMA_STUDENT']
+    destination_schema_identifier = psycopg2.sql.Identifier(destination_schema)
+    staging_schema = destination_schema + '_staging'
+    staging_schema_identifier = psycopg2.sql.Identifier(staging_schema)
```
raydavis: BOAC caching's thread-local approach was "fun", but I'm glad Nessie doesn't have to go there.

raydavis:
Forward to the next trench!

PR #105, opened 2018-06-29
NS-132 Write enrollment term feeds to Redshift

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-132

These seem closely enough linked to student profiles to go into the same schema and background job.


raydavis:
Makes sense to me.

PR #108, opened 2018-07-03
NS-140 Standardize name of Redshift Student schema

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-140

This week's theme: the damage one letter can make.


paulkerschen:
And how.

PR #109, opened 2018-07-03
NS-133 Internal schema containing ASC athletic profiles

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-133

Too many schemas? Check. No tests? Check. Outstanding TODOs? Check. But since this isn't yet a scheduled job, it's probably better to submit as is than leave it to languish on a working branch.

`student_profiles` contains precomposed JSON for quick serving.
`students` contains index columns suitable for querying on ASC attributes.

Output available for inspection under nessie_dev/boac_advising_asc_pk_test.


raydavis:
A reasonable start. Thanks, @paulkerschen .

PR #112, opened 2018-07-05
NS-131 Add uid, name columns to student index table

paulkerschen:

So many searches.


raydavis:
Every one a pip.

PR #113, opened 2018-07-06
NS-143 Make GenerateMergedStudentFeeds term-specific, schedule for current term

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-143


sandeepmjay:
LGTM @paulkerschen. Merging.

PR #114, opened 2018-07-07
NS-144 More Nessie job scheduling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-144

Timing can be tweaked via the scheduler API, but I think this is the order of operations.


raydavis:
Let's see how she runs on the highway. Thank you for taking up four armloads of slack, @paulkerschen

PR #115, opened 2018-07-09
NS-106 Extract Term GPAs for students

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-106

Not the top priority, but that's why I did it.


paulkerschen:
But _why_ was Spring 2016 such a good term for Oski?

PR #116, opened 2018-07-09
NS-146 API endpoints to trigger new jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-146


sandeepmjay:
Looks good @paulkerschen. Shipping it !

PR #118, opened 2018-07-10
NS-145 Better logging for unexpected errors in background jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-145


sandeepmjay:
Error logging FTW!

PR #120, opened 2018-07-10
NS-148 Expand student population to ASC+COE

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-148

### DIFF
--- a/nessie/lib/queries.py
+++ b/nessie/lib/queries.py
@@ -31,12 +31,26 @@ from nessie.lib.mockingdata import fixture
 data_loch_db = None
 
 
+def asc_schema():
+    return app.config['REDSHIFT_SCHEMA_ASC']
+
+
 def boac_schema():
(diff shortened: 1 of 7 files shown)


nessie/lib/queries.py
```diff
@@ -31,14 +31,28 @@
...
+
+
 def intermediate_schema():
     return app.config['REDSHIFT_SCHEMA_INTERMEDIATE']
 
 
+def get_all_student_ids():
+    sql = f"""SELECT sid FROM {asc_schema()}.students
+        UNION SELECT sid FROM {coe_schema()}.students"""
+    return redshift.fetch(sql)
```
paulkerschen: This query here is the important bit.

raydavis: I concur.

raydavis:
In union there is strength.

PR #121, opened 2018-07-11
NS-149 API endpoint to update scheduled job args

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-149


raydavis:
Admin tools view here we come!

PR #125, opened 2018-07-12
NS-149 Merged feed jobs for non-current terms are enrollments only

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-149


raydavis:
Complex, schmomplex, so long as it's readable.

PR #127, opened 2018-07-13
BOAC-996 ASC API URL transition is complete

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-996


paulkerschen:
Okay gang, let’s see who this masked man really is.

PR #129, opened 2018-07-17
NS-156 Switch SIS profile and degree progress stash from RDS JSON to S3/Redshift

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-156


tests/test_jobs/test_import_degree_progress.py
```diff
@@ -1,51 +0,0 @@
...
-"""
-
-from nessie.jobs.import_degree_progress import ImportDegreeProgress
-from nessie.models import json_cache
-from tests.util import assert_background_job_status, capture_app_logs
-
-
-class TestImportDegreeProgress:
-
-    def test_import_loop(self, app, caplog, metadata_db):
```
paulkerschen: As in #128, for now these tests are casualties of having too much Amazon fanciness to mock out.

PR #130, opened 2018-07-17
NS-155, untangle ASC schema; remove asc models and prep for RDS drops

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-155

Pull ASC feed and go directly to internal schema. No more diff-with-previous-import. Instead, we push raw-data snapshot to S3; future generations can manually diff.  If this PR is merged then I'll move on to dropping RDS tables.


tests/test_jobs/test_import_asc_athletes.py
```diff
@@ -27,43 +27,11 @@
...
-        assert inactive_student.athletics[0].group_code == 'MWP'
-        assert status is not False
-        assert (not status['warnings'])
-        counts = status['change_counts']
-        assert counts['deleted_students'] == 1
-        assert counts['deleted_memberships'] == 7
-        assert counts['new_memberships'] == 2
-        assert counts['new_team_groups'] == 1
-        assert True
-
```
johncrossman: We have no easy way to mock the Amazon `COPY tsv` performed in asc_import.

tests/test_lib/test_asc.py
```diff
@@ -1,232 +0,0 @@
...
-                'Football',
-                '2017-18',
-                'Yes',
-            ),
-        ]
-        # Run import script
-        status = asc.merge_student_athletes(asc_data, delete_students=False)
-        assert 1 == status['new_team_groups']
-        eccentric_football_group = Athletics.query.filter_by(group_code=ambiguous_group).first()
-        assert 'Football - Other' == eccentric_football_group.group_name
```
johncrossman: Removal of tests above is no great loss. They are oriented towards 'model' verification and the models are gone.

paulkerschen:
Thanks for lightening the load, @johncrossman!

PR #131, opened 2018-07-17
NS-155, drop RDS tables and related db_session mgmt

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-155

Stop, drop and delete. 

When RDS usage in `initialize_job_schedules` is gone we can drop db completely.


nessie/models/development_db.py
```diff
@@ -139,35 +89,35 @@ def load_student_athletes():
         'team_code': 'TNW',
         'team_name': 'Women\'s Tennis',
     }
-    create_student(
+    _create_student(
```
paulkerschen: @johncrossman Is it possible that this ASC test data creation can go also? There's now fake-Redshift test data for students of all types in `fixtures/loch.sql`, so it might be that these are just creating duplicate rows and I missed removing them earlier.

paulkerschen: @johncrossman Sorry, ignore the above - I was thinking about new fixtures in the BOAC codebase. Too much overlap!

johncrossman:
PR amended. `development_db.py` is gone.

paulkerschen:
Whoosh. Thanks @johncrossman.

PR #133, opened 2018-07-19
BOAC-1033 Better background job logging, failure tolerance

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1033

- Use a staging table on API loops so that today's failed retrieval doesn't wipe out yesterday's success;
- Don't log 404s as errors for API loops which return 404s in the normal course of business;
- Fix a dumb bug that caused merged profile generation to fail if degree progress was missing;
- Both succeeded and failed jobs can return a status string which will be stored in the metadata table.

### DIFF
--- a/nessie/merged/sis_profile.py
+++ b/nessie/merged/sis_profile.py
@@ -56,9 +56,8 @@ def get_merged_sis_profile(csid):
     if sis_profile['academicCareer'] == 'UGRD':
         dp_result = queries.get_sis_api_degree_progress(csid)
         degree_progress_api_feed = dp_result and dp_result[0] and json.loads(dp_result[0]['feed'])
-        if not degree_progress_api_feed:
-            return False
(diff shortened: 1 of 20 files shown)


nessie/merged/sis_profile.py
```diff
@@ -56,9 +56,8 @@ def get_merged_sis_profile(csid):
     if sis_profile['academicCareer'] == 'UGRD':
         dp_result = queries.get_sis_api_degree_progress(csid)
         degree_progress_api_feed = dp_result and dp_result[0] and json.loads(dp_result[0]['feed'])
-        if not degree_progress_api_feed:
-            return False
```
paulkerschen: The dumb bug was right here.

johncrossman:
Dumb bug gone, smart handling added. Nice.

PR #134, opened 2018-07-19
BOAC-1035 Ignore inactive memberships for otherwise active athletes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1035


paulkerschen:
This fudge doesn't really iron out the underlying problem but should bring app behavior in line with Classic BOAC.

johncrossman:
With some simplifications you must be cruel to be kind.

PR #136, opened 2018-07-20
BOAC-1027 Fix team name ordering in student summary

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1027

The simplest one to test (after Nessie reloads data, anyway).


johncrossman:
We like simple.

PR #139, opened 2018-07-23
NS-167 Add CoE students import to scheduled job

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-167


raydavis:
I expect a QA PR will be needed as well.

PR #141, opened 2018-07-24
NS-168 Generate required schemas on startup if necessary

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-168

In addition to startup check,
- Get the components of API-triggered chained jobs in alignment with scheduled jobs;
- Rename the old CreateAscSchema job (which no longer creates the ASC schema) to GenerateAscProfiles (which is what it does).

Because of the job rename, this might be a good candidate for a pre-deploy QA PR, so that we don't get a confusingly named job stuck in the SQLAlchemy store.


raydavis:
Thanks, @paulkerschen . I agree about a pre-deploy QA PR, since that will simplify the release instructions.

PR #143, opened 2018-07-24
NS-163 Backfill new students; NS-170 Duplicate enrollment row fix

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-163
https://jira.ets.berkeley.edu/jira/browse/NS-170

### DIFF
--- a/nessie/jobs/background_job.py
+++ b/nessie/jobs/background_job.py
@@ -27,8 +27,6 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """Parent class for background jobs."""
 
 
-from datetime import datetime
-import hashlib
 import os
 import re
 from threading import Thread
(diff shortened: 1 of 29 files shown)


nessie/jobs/background_job.py
```diff
@@ -38,86 +36,9 @@
...
-    template_data.update(kwargs)
-    return template_string.format(**template_data)
-
-
-def resolve_sql_template(sql_filename, **kwargs):
-    with open(app.config['BASE_DIR'] + f'/nessie/sql_templates/{sql_filename}', encoding='utf-8') as file:
-        template_string = file.read()
-    return resolve_sql_template_string(template_string, **kwargs)
-
-
```
paulkerschen: All this stuff had to get carted over to `nessie.lib.util` because of circular import problems, but I think it looks nicer in its new home.

nessie/jobs/generate_merged_student_feeds.py
```diff
@@ -48,14 +51,43 @@ class GenerateMergedStudentFeeds(BackgroundJob):
     staging_schema = destination_schema + '_staging'
     staging_schema_identifier = psycopg2.sql.Identifier(staging_schema)
 
-    def run(self, term_id=None):
-        """Loop through all records stored in the Calnet external schema and write merged student data to the internal student schema."""
-        app.logger.info(f'Starting merged profile generation job... (term_id={term_id})')
+    def run(self, term_id=None, backfill_new_students=False):
```
paulkerschen: I've locally tested the changes below to the best of my ability, but a run through our whole population seems like the only way to cover all the bases.

raydavis:
Complex (at least in diff view), but I think I've followed the basic paths.

A couple of upcoming tasks also require fill-if-not-there-already logic, by the way.

PR #145, opened 2018-07-25
NS-171 drop_external_schema seriously means it

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-171


paulkerschen:
The perfect crime.

PR #146, opened 2018-07-26
NS-176 Handle missing academic status in feed merge

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-176


raydavis:
Thanks for the data research, too, @paulkerschen

PR #148, opened 2018-07-26
NS-177 Avoid doubling up SIS Data during term transition

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-177


paulkerschen:
Thanks, @raydavis, a stitch in time.

PR #151, opened 2018-07-26
NS-106 Term GPA of 0 might indicate 0 units taken for GPA

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-106


nessie/externals/sis_student_api.py
```diff
@@ -59,14 +59,19 @@ def get_term_gpas(cs_id):
...
                 continue
-            # Ignore terms in which the student took no classes with units.
+            # Ignore terms in which the student took no classes with units. These may include future terms.
             total_units = next((u for u in registration.get('termUnits', []) if u['type']['code'] == 'Total'), None)
             if not total_units or not total_units.get('unitsTaken'):
                 continue
             term_id = registration.get('term', {}).get('id')
             gpa = registration.get('termGPA', {}).get('average')
+            term_units = registration.get('termUnits', [])
+            units_taken_for_gpa = next(tu['unitsTaken'] for tu in term_units if tu['type']['code'] == 'For GPA')
```
paulkerschen: @raydavis Perhaps wrap this in a extra parenthesis and `None`, as on line 63, to forestall `StopIteration` errors?

raydavis: I wrote it that way at first and then thought, "But what could go wrong?", forgetting that the answer to that is always "Everything." Will amend.

PR #159, opened 2018-07-31
NS-180 Enrolled term credits can be earned before "matriculation"

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-180


paulkerschen:
Thanks @raydavis. When this code runs out in 2100 I bet our inheritors will be worrying about other things.

PR #160, opened 2018-07-31
NS-185 Fix duplicate majors, RDS refresh, transaction handling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-185


raydavis:
Makes sense in El Cerrito.

PR #162, opened 2018-08-01
NS-187 Nessie should provide any midterms/drops data it has on hand

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-187

Where "on hand" means "available from a fast SQL query rather than a slow API loop."

The client can decide whether to show it or not. BOAC appears to be up to the responsibility.


raydavis:
Travis should be happier now. (I forgot to push my local change before opening the PR.)

PR #164, opened 2018-08-03
NS-190 Diagnostic debug logging

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-190

Noisy logs! Hopefully this gives us a good spread of timepoints.


johncrossman:
Looks good. Looks right.

PR #166, opened 2018-08-03
NS-190 Index the assignment_submissions_scores user ID which is queried

raydavis:
... rather than the one which is not.

Just another contributor to https://jira.ets.berkeley.edu/jira/browse/NS-190


paulkerschen:
I'm a believer, interleaver.

PR #168, opened 2018-08-06
NS-193 Add api/schedule/reload API to restore configured schedule

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-193


johncrossman:
Thanks for test coverage, @raydavis

PR #169, opened 2018-08-07
NS-197 Merged feeds job optimization: fewer, better Redshift queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-197

- Instead of making many small queries per student per enrollment (a holdover from our API-looping patterns), make fewer large queries and pass structures around in memory to reduce the number of Redshift roundtrips.
- Instead of having the get_submissions_turned_in_relative_to_user query calculate submission counts on the fly, pull precalculated totals from a new derived table in boac_analytics.
- Keep track of previous query results for Canvas sitewide statistics to avoid requerying during the loop. (This one may not yield that much benefit, but it's in the spirit.)

Local testing suggests an order-of-magnitude performance improvement, but EC2 job runs will give the whole story.


johncrossman:
Always buy in bulk! Time for EC2 to tell its story.

PR #170, opened 2018-08-08
NS-197 No queries needed on empty course id lists

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-197

Follow-up to #169.


johncrossman:
Not never anti-unmapped!

PR #171, opened 2018-08-08
NS-189 Better transaction handling across RDS/Redshift

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-189

The point of this added object layer is that psycopg2 transactions only work if you hold on to a consistent cursor.


johncrossman:
Better handling indeed. Thanks @paulkerschen

PR #172, opened 2018-08-09
NS-98 New COE attributes for the loch

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-98


johncrossman:
Prep for the PREP.

PR #177, opened 2018-08-15
NS-202 Handle data surprises of all sorts in background jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-202


johncrossman:
Known unknowns handled.

PR #179, opened 2018-08-16
NS-205 Guard against duplication in Canvas Data sources

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-205

While we prefer GROUP BY to DISTINCT in Redshift Spectrum queries, it's not possible here because of dynamic columns set by CASE statements. Fortunately, a test run suggests we aren't taking a huge performance hit.


sandeepmjay:
Mitosis checks by Nessie!

PR #180, opened 2018-08-16
[QA] NS-205 Guard against duplication in Canvas Data sources

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-205

QA version of #179.


sandeepmjay:
Mitosis checks by Nessie!

PR #182, opened 2018-08-16
[QA] NS-203 One more try at malformed Hub feed handling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-203

QA version of #181.


sandeepmjay:
Nice work taming the Hub feeds!

PR #183, opened 2018-08-17
NS-142 Canvas enrollments API poller

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-142

A lot of old BOAC code reappears in Nessie, wrapped in a background job that loops through all sites in which our population is enrolled and stashes the raw feed together with an extracted last_activity value.


johncrossman:
Resurrect the dead (code)! Well done.

PR #184, opened 2018-08-20
NS-142 Incorporate API-derived activity timestamp when appropriate

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-142


johncrossman:
We want the BESTEST.

PR #187, opened 2018-08-21
NS-206 SIS terms API external connection + fixture

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-206


johncrossman:
Nessie is becoming a very smart girl.

PR #192, opened 2018-09-04
NS-213 Fix return value for rds.execute

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-213


raydavis:
Have you ever heard your mommy say "Noise annoys"?

PR #194, opened 2018-09-05
NS-215 Separate out schedule initialization from schedule reload

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-215

Scheduler setup code should run only during app startup.


raydavis:
It's hard to run while you're tying your shoelaces.

PR #197, opened 2018-09-07
[QA] NS-216 Another trivial change reverting previous trivial change

paulkerschen:
More pipeline testing.

https://jira.ets.berkeley.edu/jira/browse/NS-216


raydavis:
XYZ is much cooler.

PR #201, opened 2018-09-12
NS-109, Flask 1.0.2 for Nessie

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-109

In step with BOAC: https://github.com/ets-berkeley-edu/boac/pull/873


paulkerschen:
It _looks_ safe!

PR #205, opened 2018-09-17
NS-164 Let Nessie "master" & "worker" deployments share a config file

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-164

Given the choice between "Rails-y magic string conventions" and "Pythonic explicit configurations", I picked the option which requires less AWS EBS maintenance. Our new shared config filenames would therefore be EB_ENVIRONMENT minus any "-worker" or "-master" substring.

### DIFF
--- a/.ebextensions/20_download_local_configuration.config
+++ b/.ebextensions/20_download_local_configuration.config
@@ -5,7 +5,8 @@
 container_commands:
   01_get_configuration_file:
     command: |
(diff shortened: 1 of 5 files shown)


paulkerschen:
Works for me, @raydavis! Shall we put a jobless nessie-dev config file in place before merge?

raydavis:
We shalled, @paulkerschen . I'll update the release instructions, too.

PR #206, opened 2018-09-18
NS-224 Brain transplant for Mockingbird: httpretty -> responses

paulkerschen:

(For background on the responses package, see https://cra.mr/2014/05/20/mocking-requests-with-responses - it's designed after httpretty but simpler, which fits our usage and seems to cause fewer conflicts.)

### DIFF
--- a/config/testext.py
+++ b/config/testext.py
@@ -11,10 +11,14 @@ REDSHIFT_USER = 'username'
 
 REDSHIFT_SCHEMA_ASC = 'testext_mynamehere_asc'
 REDSHIFT_SCHEMA_BOAC = 'testext_mynamehere_boac'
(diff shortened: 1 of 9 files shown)


config/testext.py
```diff
@@ -11,10 +11,14 @@
 
 REDSHIFT_SCHEMA_ASC = 'testext_mynamehere_asc'
 REDSHIFT_SCHEMA_BOAC = 'testext_mynamehere_boac'
+REDSHIFT_SCHEMA_CALNET = 'testext_mynamehere_calent'
```
raydavis: tpyo?

paulkerschen: Ayup.

raydavis:
Cleanliness is next to responsiveness.

PR #207, opened 2018-09-19
NS-224 Make ImportTermGpas testable as model for other jobs

paulkerschen:

Test data features recurring characters familiar to fans of the BOAC codebase.


raydavis:
You can never have too many tests or too much Preston Sturges.

PR #208, opened 2018-09-19
NS-224 Get rid of preposterous JSON-into-CSV fixtures

paulkerschen:

They were never meant to be.


johncrossman:
There were never a part of His plan.

PR #210, opened 2018-09-20
BOAC-1259 Include comparative percentiles for course means

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1259

Follow-up to #209. For matrix plotting we need comparative percentiles, not just raw values.


johncrossman:
The zptile and zscore look zcorrect.

PR #211, opened 2018-09-24
BOAC-1250 One Nessie can handle multiple BOAC refreshes

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1250

If this looks OK, we'll need to pause before merging so as to make the necessary edit to Dev configuration.

### DIFF
--- a/nessie/externals/boac.py
+++ b/nessie/externals/boac.py
@@ -31,11 +31,14 @@ from nessie.lib import http
 
 
 def kickoff_refresh():
-    return authorized_request(app.config['BOAC_CACHE_REFRESH_URL'])
+    successful = True
+    for boac in app.config['BOAC_REFRESHERS']:
+        successful = authorized_request(boac['URL'], boac['API_KEY']) and successful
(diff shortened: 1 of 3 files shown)


nessie/externals/boac.py
```diff
@@ -31,11 +31,14 @@
 
 
 def kickoff_refresh():
-    return authorized_request(app.config['BOAC_CACHE_REFRESH_URL'])
+    successful = True
+    for boac in app.config['BOAC_REFRESHERS']:
+        successful = authorized_request(boac['URL'], boac['API_KEY']) and successful
```
johncrossman: @raydavis Should we log a warning if request fails?

raydavis: Thanks, @johncrossman . I thought underlying code took care of that, but I'll double check.

raydavis: Confirmed, @johncrossman -- a failing response is logged at ERROR.

PR #214, opened 2018-09-26
BOAC-1295 Zeros only where they ought to be

paulkerschen:


raydavis:
Nothing matters except when it doesn't.

PR #215, opened 2018-09-27
[QA] BOAC-1295 Zeros only where they ought to be

paulkerschen:

QA version of #214.


sandeepmjay:
It's all about zeroes!

PR #216, opened 2018-09-27
[QA] BOAC-1295 Zeros only where they ought to be

paulkerschen:

Version of #215 actually against QA this time, bah.


raydavis:
I blame society.

PR #217, opened 2018-10-02
LA-521 Utility methods for interacting with DMS endpoints and tasks

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/LA-521


johncrossman:
Come on in the data lake, DMS. The water's fine.

PR #219, opened 2018-10-04
NS-240 Nessie and the LRS can be friends

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-240


johncrossman:
Aww, they're playing so nicely!

PR #223, opened 2018-10-05
NS-239 Scheduling for LRS incremental import job

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-239

Adding the job to the scheduler as paused will allow us to turn it on via the API as desired.

Also, may as well remove the confusing arbitrary job times from default configs.


config/default.py
```diff
@@ -84,16 +84,17 @@
 JOB_SCHEDULING_ENABLED = True
 
 # See http://apscheduler.readthedocs.io/en/latest/modules/triggers/cron.html for supported schedule formats.
```
raydavis: Since the URL link doesn't mention Python dictionaries, would it be OK to include an "e.g." sample value? The existing fill-ins were annoying noise so far as runtime utility goes, but they helped meet the default config's documentary goal.

paulkerschen: Oh yes, sample value in a comment is a good idea.

paulkerschen: Done.

raydavis:
Thanks, @paulkerschen !

PR #225, opened 2018-10-08
NS-255 Enhacements to LRS import pipeline

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-255

- Folder nesting in destination buckets; 
- Whether to truncate the source LRS Aurora is now specificable and by default turned off in the API.


johncrossman:
No problem with transients here. Thanks @paulkerschen

PR #227, opened 2018-10-09
NS-260 Caliper import pipeline: unload statements from Redshift to ETL

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-260


johncrossman:
Locked and unloaded.

PR #228, opened 2018-10-10
NS-235 WORKER_QUEUE_ENABLED should only affect Worker instances

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-235

This should let us enable background-thread pools on Nessie-worker instances while letting Nessie-master proceed By All Threads Necessary.


paulkerschen:
Labor, get in line.

PR #229, opened 2018-10-11
NS-262 Selective refresh for term GPA index in RDS

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-262


raydavis:
Last In, All Out.

PR #232, opened 2018-10-11
[NS-263] Logic to delete glue jobs via an api call

sandeepmjay:
Background job to delete LRS related glue jobs via an API call. 

https://jira.ets.berkeley.edu/jira/browse/NS-263


paulkerschen:
Thanks @sandeepmjay.

PR #235, opened 2018-10-16
[NS-243] Logic to run glue jobs to transform and monitor LRS canavs caliper data into flat structures.

sandeepmjay:
Background job to trigger exiting glue job that transforms caliper data. the job also monitors for the progress of the job and perform suitable error handling. One the job is complete the flatten version of canvas callper data is availabel at the GLUE export locations in the transient bucket. 
https://jira.ets.berkeley.edu/jira/browse/NS-243


nessie/jobs/transform_lrs_incrementals.py
```diff
@@ -26,13 +26,86 @@
...
+
+        response = glue.start_glue_job(
+            self.job_name, job_arguments,
+            app.config['LRS_CANVAS_GLUE_JOB_CAPACITY'],
+            app.config['LRS_CANVAS_GLUE_JOB_TIMEOUT'],
+        )
+        if not response:
+            app.logger.error('Failed to create Glue job')
+            return False
+        elif response['JobRunId']:
```
johncrossman: @sandeepmjay  would this line be better as `if 'JobRunId' in response:`

nessie/jobs/transform_lrs_incrementals.py
```diff
@@ -26,13 +26,86 @@
...
+                    return False
+                elif response['JobRun']['JobRunState'] == 'SUCCEEDED':
+                    app.logger.info(f'Caliper glue transformation job completed successfully: {response}')
+                    break
+                elif response['JobRun']['JobRunState'] == 'FAILED' or response['JobRun']['JobRunState'] == 'TIMEOUT':
+                    app.logger.error(f'Caliper glue transformation job failed or timed out: {response}')
+                    return False
+
+                sleep(30)
+            return True
```
johncrossman: The `while True` loop above has both `break` (ie, return True) and `return False`.  I suggest:
1. replace `break` with `return True`
1. OR, set a `result` variable and use `break` in all the if-blocks

johncrossman:
Looks good. Two minor questions.

PR #241, opened 2018-10-23
[NS-261] Canvas Caliper flattened template from the glue process.

sandeepmjay:
Flattened Caliper canvas feeds structure for redshift spectrum. Part of the JIRA

https://jira.ets.berkeley.edu/jira/browse/NS-261


paulkerschen:
Thanks @sandeepmjay!

PR #243, opened 2018-10-23
NS-273 Add non-datestamped transient Redshift unload to Caliper import pipeline

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-273

Per discussion with @sandeepmjay, it's this unload to the transient bucket that will be used as an input to Glue, not the datestamped unloads in environment-specific buckets.


sandeepmjay:
Thanks @paulkerschen!

PR #249, opened 2018-10-26
NS-230, Nessie console shows 'Failures from last sync' and 'Background Jobs of mm/dd/yyyy'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-230

UX still needs work but, for now, I'm concerned with surfacing useful info (eg, job status).

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,5 @@ install:
   - pip3 install tox
   - npm install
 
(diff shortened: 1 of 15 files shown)


raydavis:
Better to know than to wonder. Thank you, @johncrossman

PR #250, opened 2018-10-26
NS-274 Chained Caliper import available through API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-274


sandeepmjay:
Looks good. Waiting for Mr.T.

PR #253, opened 2018-10-30
NS-232, improve vue-router setup prior to CAS activation in nessie-dev

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-267


src/router.ts
```diff
@@ -1,47 +1,70 @@
 import Vue from 'vue';
```
johncrossman: The diff below is easier to read with `?w=1` appended to github URL.

PR #255, opened 2018-10-30
NS-286 Nightly scheduling for Caliper load pipeline

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-286

Three of our four new jobs are about moving files around and belong together in a chain. The fourth is about building Redshift views and needs to be slotted into our table-building hierarchy at the appropriate place.


nessie/jobs/generate_canvas_caliper_analytics.py
```diff
@@ -56,18 +55,10 @@ def run(self):
...
                 app.logger.info('Generating user request tables and caliper analytics.')
-                redshift_schema_canvas = app.config['REDSHIFT_SCHEMA_CANVAS']
-                redshift_schema_caliper_analytics = app.config['REDSHIFT_SCHEMA_CALIPER']
 
-                resolved_ddl_caliper_analytics = resolve_sql_template(
-                    'generate_caliper_analytics.template.sql',
-                    redshift_schema_lrs_external=redshift_schema_lrs_external,
-                    redshift_schema_canvas=redshift_schema_canvas,
-                    redshift_schema_caliper_analytics=redshift_schema_caliper_analytics,
-                )
```
paulkerschen: Note that these values don't need to be parameterized since they're provided as defaults in nessie/lib/util.py, below. Elsewhere in this file, `resolve_sql_template` takes parameters that really do vary and continue to be passed in from code.

nessie/jobs/generate_canvas_caliper_analytics.py
```diff
@@ -56,18 +55,10 @@ def run(self):
...
+                resolved_ddl_caliper_analytics = resolve_sql_template('generate_caliper_analytics.template.sql')
                 if redshift.execute_ddl_script(resolved_ddl_caliper_analytics):
-                    app.logger.info('Caliper analytics tables successfully created.')
-                    return True
+                    return 'Caliper analytics tables successfully created.'
```
paulkerschen: Returning a string rather than `True` ensures the string will end up in the metadata.background_job_status table.

sandeepmjay:
Looks good @paulkerschen. Thanks for the notes on the changes.

PR #257, opened 2018-10-31
NS-281 Build S3-bound TSVs as binary strings for smaller memory footprint

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-281


johncrossman:
More of that good Python shorthand.

PR #258, opened 2018-10-31
NS-231, kick off jobs, with required args, from Nessie admin console

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-231

If new create-job api endpoints are created on the backend then the option-to-run should automatically surface on the console with "input required args", if necessary.


src/router.ts
```diff
@@ -1,13 +1,30 @@
...
 
 Vue.use(VueRouter);
 
+let registerMe = () => {
+  return Promise.resolve(
+    getMyProfile().then(me => {
+      if (me) {
+        store.commit('registerMe', me);
+        getRunnableJobs().then(data => {
+          store.commit('cacheRunnableJobs', data);
```
johncrossman: Runnable-job options are cached for entire user session.

paulkerschen:
Thanks @johncrossman - it's looking nice and orderly indeed.

raydavis:
I look forward to the upcoming SFTP-to-AWS-S3 client!

PR #261, opened 2018-11-01
NS-287 Students not on CoE & ASC lists should be removed from Nessie

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-287

This is one ugly tangle of code, but I didn't want to get into a full-out refactoring to separate the interwoven strands of term-specific and non-term-specific data updates. Let me know if I'm being too timid.


paulkerschen:
It's tangled all right, but perhaps best sorted out incrementally. Thanks @raydavis.

PR #263, opened 2018-11-01
BOAC-1272 Calculate against unique scores for optimal matrix spread

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1272

This is functionally Paul Kerschen's https://github.com/paulkerschen/nessie/commit/458f2bfe5133a31e51eed83c5715d53933b50213

My only changes were to keep the old percentile property alongside the new one, and to add a display percentile to the course mean.

I'll wait to submit the corresponding BOAC PR until after this gets a chance to run.

PR #264, opened 2018-11-02
BOAC-1272 Change incorrect Mean Man display percentile

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1272

Correction to https://github.com/ets-berkeley-edu/nessie/pull/263 , which sometimes claimed that the Mean Man was in the "30th" percentile, etc.


paulkerschen:
What is "correct" when it comes to the Mean Man? Time will tell.

PR #266, opened 2018-11-05
BOAC-1272 Match the Mean Man to middling students

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1272

"This time FOR SURE!"


johncrossman:
Bullwinkle, that trick never works!

PR #267, opened 2018-11-05
[NS-297] Included logic to fix course_id parsing anomalies.

sandeepmjay:
https://jira.ets.berkeley.edu/jira/browse/NS-297


paulkerschen:
Thanks @sandeepmjay!

PR #268, opened 2018-11-06
NS-291 Add more CoE student attributes

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-291

### DIFF
--- a/fixtures/students.sql
+++ b/fixtures/students.sql
@@ -26,11 +26,17 @@ CREATE TABLE {redshift_schema_coe}.students
     advisor_ldap_uid VARCHAR NOT NULL,
     gender VARCHAR NOT NULL,
     ethnicity VARCHAR NOT NULL,
-    minority VARCHAR NOT NULL,
-    did_prep VARCHAR NOT NULL,
-    prep_eligible VARCHAR NOT NULL,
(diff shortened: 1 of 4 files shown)


johncrossman:
Looks good, @raydavis  Our pesky linter wants a comma.

paulkerschen:
@raydavis It looks like the test-only version of this table (defined in `/fixtures/students.sql`) isn't raising a fuss about missing the new columns because our test coverage is still so spotty. But in hopes of someday, somehow filling these tests out, should we have the fixture keep pace?

raydavis:
Thanks, @johncrossman . As often happens, I fixed `tox` locally and then forgot to push the fix to GitHub.

@paulkerschen , I noticed that `{redshift_schema_coe}.students` fixture SQL seemed to define VARCHAR column types where the real derived-table has Boolean. Was that intended?

paulkerschen:
@raydavis Not intended! Probably that fixture was erroneously copied from the external rather than the internal schema.

raydavis:
Thank, @paulkerschen , I'll amend while I'm in there.

raydavis:
PR amended, @paulkerschen and/or @johncrossman

PR #270, opened 2018-11-06
NS-231, kick off new Nessie job; required params in path, if necessary

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-231

### DIFF
--- a/src/components/Jobs.vue
+++ b/src/components/Jobs.vue
@@ -14,7 +14,10 @@
             {{requiredParameter}}: <input v-model="runnableJob.params[index]" placeholder="Required">
           </div>
         </div>
-        <button @click="runSelectedJob" :disabled="!runnableJob.ready()">Run it!</button>
+        <div>
+          <button @click="runSelectedJob" :disabled="!runnableJob.ready()" v-if="!runnableJob.isRunning">Run it!</button>
+          <span v-if="runnableJob.isRunning">Running...</span>
+        </div>
       </div>
       <div>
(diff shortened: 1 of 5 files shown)


src/components/Jobs.vue
```diff
@@ -92,7 +96,21 @@ export default {
       });
     },
     runSelectedJob() {
-      console.log('Run job: ' + this.runnableJob.selected.name);
+      let apiPath = this.runnableJob.selected.path;
```
paulkerschen: @johncrossman Idle curiosity - why `let` here? I know there are subtle effects on scoping but am not used to the idioms.

src/components/Jobs.vue
```diff
@@ -92,7 +96,21 @@ export default {
...
+      this._.each(
+        this.runnableJob.selected.requiredParameters,
+        (requiredParameter, index) => {
+          apiPath = this._.replace(
```
paulkerschen: So `this._` is the usual way to inject Lodash into Vue?

johncrossman: Yes, a big odd – see https://github.com/Ewocker/vue-lodash  There might be a better util and/or usage.

src/components/Jobs.vue
johncrossman: `let` is preferred ES6 usage. See https://vegibit.com/es6-let-vs-var-vs-const/

paulkerschen:
Thanks @johncrossman!

PR #272, opened 2018-11-06
NS-299, 'npm run build-vue' failed due to improper vue-lodash import

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-299

`npm run build-vue` has stricter standards compared to `npm run serve` (used for local dev). The inline "eslint:warn" will be addressed later.


paulkerschen:
Ah, clever, and we lose that weird `this` in the bargain. Waiting on T.

PR #279, opened 2018-11-19
NOJIRA Replace some TypeScript punts; fix linting

raydavis:
Following up on a discussion with @johncrossman & @lyttam (which would've also included @paulkerschen if he'd been available)...

Nessie/BOA's wee bit of TypeScript code includes some explicit function-parameter typing of "any" even though the code clearly does expect an object with a predictable structure.

None of us are eager to relive the full-on interface/class Java experience. But the newer JS frameworks push TS hard, and if we're going to use it, it would be nice to trim the noise. So we'll try this:

* Fill in more real-types when it's easy (e.g., this PR).
* Instead of explicitly defaulting all arguments to "any", we will leave them explicitly _untyped_, which in turn means continuing to disable "noImplicitAny" checks. That way it'll be easy to see which typing has been thought out by a developer and which typing hasn't been handled yet.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import VueRouter from 'vue-router';
+import VueRouter, { Route } from 'vue-router';
 import store from '@/store';
 import { getRunnableJobs } from '@/api/job';
 import { getMyProfile } from '@/api/user';
@@ -25,8 +25,8 @@ let registerMe = () => {
   );
 };
 
-let beforeEach = (to: any, from: any, next: any) => {
-  let safeNext = (to: any, next: any) => {
+let beforeEach = (to: Route, from: Route, next: Function) => {
+  let safeNext = (to: Route, next: Function) => {
     if (to.matched.length) {
       next();
     } else {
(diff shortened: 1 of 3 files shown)


src/router.ts
```diff
@@ -1,5 +1,5 @@
 import Vue from 'vue';
-import VueRouter from 'vue-router';
+import VueRouter, { Route } from 'vue-router';
```
raydavis: Our TypeScript linter(s) threw a bogus "defined but never used (no-unused-vars)" error on the Route interface despite its use below. "You had ONE job..." After following the issue-report bounces across three different GitHub projects, I tried the following upgrades to no avail:

```
-    "@vue/eslint-config-typescript": "^3.0.5",
+    "@vue/eslint-config-typescript": "^3.1.1",
+    "eslint-plugin-typescript": "^0.13.0",
+    "typescript-eslint-parser": "^21.0.1",
```

paulkerschen:
Thanks, Team Typescript! This looks sensible to me.

johncrossman:
+1 @raydavis

PR #280, opened 2018-11-20
COL-1367 Add conversation message tables to Canvas external schema

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1367


johncrossman:
Those external tables are a dime a dozen.

PR #282, opened 2018-11-27
NS-303, explicit types in ts files; Travis verifies that ts compiles

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-303

Added check to Travis build so we don't fail far along the pipeline.


paulkerschen:
Thanks for keeping Travis honest, @johncrossman.

PR #288, opened 2018-12-06
NS-306 Replace Enrollments API with Junction enrollments data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-306

PLEASE REVIEW BUT DO NOT MERGE UNTIL PREP WORK IS DONE.

Currently, the required Junction exports can only be produced from Junction-Dev. The new Junction code must make its way to production before this change can safely reach Nessie-Prod.

Also currently, Nessie-Dev gets its Junction exports from the production run. Junction-Dev will need to start exporting to a non-prod location, and Nessie-Dev will need reconfiguring before this change will run successfully.


paulkerschen:
This looks good to me, @raydavis, and all those removed lines of code + fixtures are easy on the eyes.

As mentioned over in https://jira.ets.berkeley.edu/jira/browse/BOAC-1298, as a follow-up I hope this might enable us to reduce the complexity of the backfilled-terms logic over in generate_merged_student_feeds.py.

raydavis:
Sure enough, that run had a failure. I'll create a follow-up JIRA, but I don't think it should block this PR.

raydavis:
@johncrossman @paulkerschen : Prep steps are done and this is ready for merging (unless we have a veto).

PR #292, opened 2018-12-14
NS-314 Avoid duplicates in sis_dropped_classes table

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-314

### DIFF
--- a/nessie/sql_templates/create_intermediate_schema.template.sql
+++ b/nessie/sql_templates/create_intermediate_schema.template.sql
@@ -81,6 +81,10 @@ AS (
             en.enrollment_status = 'D'
             OR en.grade = 'W'
         )
+    /* Clear out duplicates. */
+    GROUP BY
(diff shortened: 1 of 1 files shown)


nessie/sql_templates/create_intermediate_schema.template.sql
```diff
@@ -81,6 +81,10 @@ AS (
             en.enrollment_status = 'D'
             OR en.grade = 'W'
         )
+    /* Clear out duplicates. */
+    GROUP BY
```
sandeepmjay: @raydavis - Would we need a having clause?

raydavis: Not that I can see off-hand, @sandeepmjay -- please say more!

sandeepmjay: Looks like this was a filtering query. Thanks for the explanation @raydavis.

sandeepmjay:
LGTM @raydavis.

PR #300, opened 2019-01-07
NS-331 Avoid scheduling conflicts between the final two BOA jobs

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-331

Required flipping GenerateMergedStudentFeeds defaults to match the nightly job.

Since these scheduling conflicts have been hitting production for a while, if this change tests out I think it should probably go into the QA branch as well.

### DIFF
--- a/nessie/api/job_controller.py
+++ b/nessie/api/job_controller.py
@@ -134,8 +134,6 @@ def generate_intermediate_tables():
 @app.route('/api/job/generate_merged_student_feeds/<term_id>', methods=['POST'])
 @auth_required
 def generate_merged_student_feeds(term_id):
-    if term_id == 'all':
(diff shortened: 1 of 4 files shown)


raydavis:
Travis OKs this now. I wasn't able to reproduce the test failure locally until I manually dropped and recreated the test DBs.

paulkerschen:
Thanks, @raydavis, makes sense. As with the persistence of the old job in your test environment, this PR may need a post-deploy release instruction to manually delete the Postgres row representing JOB_REFRESH_BOAC_CACHE from the SQLAlchemy job store in Nessie RDS. Otherwise that job, even if unscheduled, will persist as noise.

PR #310, opened 2019-01-16
NS-340: falls back on UCBX academic status if no other status

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-340

I think this is the right thing to do for now, but there may need to be a broader discussion about what to do with students who have multiple academic statuses.


paulkerschen:
Thanks for fixture and all, @lyttam; looks right to me.

PR #315, opened 2019-01-17
NS-347 Disambiguate column names in IndexEnrollments GROUP BY

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-347


lyttam:
Thanks for the fix.

PR #316, opened 2019-01-18
[NS-349] Background job and api to Catalog Edx external schemas on Nessie

sandeepmjay:
Create and restore external schemas on transformed Edx data on S3 buckets for BerkeleyX tenants
https://jira.ets.berkeley.edu/jira/browse/NS-349


paulkerschen:
Thanks @sandeepmjay, looks good!

PR #317, opened 2019-01-22
NS-344: adds min and max units to student profile

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-344


johncrossman:
Thank you for the test coverage, @lyttam

PR #320, opened 2019-01-24
COL-1374 Nessie swims into the world of SuiteC data management

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/COL-1374

After some thrashing, I've ended up with a script intended to be run monthly, which uses a date-based cutoff for events instead of truncating the entire table, and thus doesn't require the app to be paused.


johncrossman:
That's some smashing BASHing, @paulkerschen

PR #321, opened 2019-01-24
NS-316 Fix front-end error when running Nessie jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-316

This doesn't yet allow us to run jobs with args, but one step at a time.


johncrossman:
We keep on truckin'

PR #322, opened 2019-01-25
NS-353 Nessie-side support for Physics department advising

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-353

This work went faster than expected since we have no physics-department-specific profile information to merge in; it's just a list of SIDs.

For now CreatePhysicsSchema is set up as a manual job only, since we aren't yet expecting regular updates in the source data.

Includes a bit of config rearrangement in an attempt at noise reduction.


johncrossman:
You can't fight it. It's Physics.

PR #324, opened 2019-01-28
 NS-319 Batch GenerateMergedStudentFeeds queries for performance

raydavis:

Reviewers might want to start with `queries.py` and then move to `generate_merged_student_feeds.py` and `student_terms.py`.

I'll create a follow-up JIRA for the _next_ speed-up attempt. In particular it would be nice to find some way to reduce the costs of the submissions-comparison query and the final analytics calculation loop.


raydavis:
I got smacked by the same lint-py change that got @paulkerschen earlier today, but everyone seems to be getting along now.

paulkerschen:
@raydavis Having contributed to so much of this tangle, I appreciate your fortitude in parceling it back out. Let it fly!

PR #325, opened 2019-01-30
NS-319 Revert PR #324 while we deal with the AWS-side query issue

raydavis:

See the last comment at JIRA for background on this.


johncrossman:
A couple py linter errors. Ready to merge when fixed.

PR #327, opened 2019-02-01
NS-319 Another try at the big merged feeds speed-up

raydavis:

This time we hope it will be both fast _and_ non-fatal!

Also filters out bad rows from the existing assignment_submissions_scores query. If you want to just see the new stuff, check the second commit.


paulkerschen:
Down the hatch!

PR #328, opened 2019-02-04
NS-221 Track enrollment status changes for current-term primary sections

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-221

### DIFF
--- a/nessie/externals/redshift.py
+++ b/nessie/externals/redshift.py
@@ -60,12 +60,16 @@ def execute_ddl_script(sql):
     statements = sql.split(';')
     # Remove any trailing debris after the last SQL statement.
     del statements[-1]
-    for index, statement in enumerate(statements):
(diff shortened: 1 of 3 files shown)


nessie/externals/redshift.py
```diff
@@ -60,12 +60,16 @@ def execute_ddl_script(sql):
     statements = sql.split(';')
     # Remove any trailing debris after the last SQL statement.
     del statements[-1]
-    for index, statement in enumerate(statements):
-        app.logger.info(f'Executing DDL script {index + 1} of {len(statements)}')
-        result = execute(statement)
-        if not result:
-            app.logger.error('Error executing statement from DDL script; aborting remainder of script.')
+    with _get_cursor() as cursor:
```
paulkerschen: Using a persistent cursor/session when executing DDL scripts lets temp tables hang around.

lyttam: Nice, thanks for the update.

PR #329, opened 2019-02-04
NS-325: Maps theoretical filename to actual one

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-325


johncrossman:
Putting theory into practice. Thanks @lyttam

PR #330, opened 2019-02-05
NS-369, Nessie console: surface config, ping status, version

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-369

And, minor `store` restructuring to match (what seems to work in) BOAC.


raydavis:
Thanks for giving Nessie a little attention, @johncrossman

PR #331, opened 2019-02-05
NS-309 Rewrite GenerateCanvasCaliperAnalytics for performance

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-309

Transform bite-size daily chunks of the Caliper feed instead of retransforming all of human history every night.

### DIFF
--- a/nessie/jobs/generate_canvas_caliper_analytics.py
+++ b/nessie/jobs/generate_canvas_caliper_analytics.py
@@ -23,24 +23,48 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
+from datetime import datetime, timedelta
+import os
+
 from flask import current_app as app
-from nessie.externals import redshift
+from nessie.externals import redshift, s3
 from nessie.jobs.background_job import BackgroundJob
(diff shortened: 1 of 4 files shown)


raydavis:
Looks great to me, @paulkerschen , but I'll give @sandeepmjay a few hours to check it, too.

nessie/jobs/generate_canvas_caliper_analytics.py
```diff
@@ -49,21 +73,36 @@ def run(self):
             app.logger.error('Caliper explode schema and table creation failed.')
             return False
 
-        app.logger.info('Verify if data exists in caliper explode tables')
-        redshift_response = redshift.fetch(f'select count(*) from {redshift_schema_lrs_external}.{canvas_caliper_explode_table}')
+        # Sanity-check event times from the latest Caliper batch against previously transformed event times.
+        def datetime_from_query(query):
```
lyttam: Why define this function inside of `run` instead of at the class level?

paulkerschen: It's similar to defining a proc or lambda in Ruby, except that Python functions-within-functions let you give it a more descriptive name. The convention with these nested functions seems to be a signal that you only expect it to be useful for the code that immediately follows, and not to be more generally applicable.

lyttam: Ah, makes sense. Thanks for the explain.

PR #333, opened 2019-02-07
NS-309 Caliper transformation: adjust tolerance for timestamp sanity check

paulkerschen:
News flash, kid: in the real world your timestamps don't always arrive in perfect order.

https://jira.ets.berkeley.edu/jira/browse/NS-309


johncrossman:
I read the news, oh boy!

PR #334, opened 2019-02-07
NS-371: Provides SIS advising notes internal schema for boac

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-371


paulkerschen:
It's data!

PR #335, opened 2019-02-07
NS-373 Disambiguate RDS and Redshift schemas

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-373


johncrossman:
Say what you mean and mean what you say.

PR #336, opened 2019-02-07
NS-371: Fixes duplicate note rows

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-371

### DIFF
--- a/nessie/sql_templates/create_sis_advising_notes_schema.template.sql
+++ b/nessie/sql_templates/create_sis_advising_notes_schema.template.sql
@@ -303,12 +303,13 @@ AS (
     ON N.sid = D.sid
     AND N.institution = D.institution
     AND N.note_id = D.note_id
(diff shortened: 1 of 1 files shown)


nessie/sql_templates/create_sis_advising_notes_schema.template.sql
```diff
@@ -308,7 +308,8 @@ AS (
     ON N.note_category = C.note_category
     JOIN
         {redshift_schema_sis_advising_notes}.advising_note_subcategories S
-    ON N.note_subcategory = S.note_subcategory
+    ON N.note_category = S.note_category
+    AND N.note_subcategory = S.note_subcategory
```
johncrossman: @lyttam I still see duplicates w/ the `subcategory` condition above. What about a DISTINCT constraint?

PR #337, opened 2019-02-07
NS-374: Adds criteria to select only latest version of a note

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-374


johncrossman:
We serve only the freshest notes.

PR #338, opened 2019-02-07
NS-304 Add external table for SIS-provided advisee demographics

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-304

(This should not be merged until after the QA branch is cut, since it embeds a next-release-Junction dependency.)


paulkerschen:
Multidimensional demographics, here we come.

PR #339, opened 2019-02-08
NS-372 Raise explicit errors in background jobs where foreseeable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-372

### DIFF
--- a/tests/test_jobs/test_import_asc_athletes.py
+++ b/tests/test_jobs/test_import_asc_athletes.py
@@ -26,8 +26,10 @@ ENHANCEMENTS, OR MODIFICATIONS.
 import json
 
 from nessie.externals import asc_athletes_api
+from nessie.jobs.background_job import BackgroundJobError
 from nessie.jobs.import_asc_athletes import ImportAscAthletes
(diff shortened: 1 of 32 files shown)


tests/test_jobs/test_import_asc_athletes.py
```diff
@@ -51,8 +53,8 @@ def test_update_safety_check(self, app):
         }
         modified_response = MockResponse(200, {}, json.dumps(skinny_import))
         with register_mock(asc_athletes_api._get_asc_feed_response, modified_response):
-            status = ImportAscAthletes().run()
-            assert status is False
+            with pytest.raises(BackgroundJobError):
+                ImportAscAthletes().run()
```
lyttam: @paulkerschen should there be an assertion here? Same below.

paulkerschen: @alyttle In this case `pytest.raises` is acting as an assertion; the test will fail if it doesn’t hit an exception of that class.

https://docs.pytest.org/en/latest/assert.html#assertraises

lyttam: Makes sense!

PR #340, opened 2019-02-08
NS-375: Adds primary key to advising_notes

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-375


paulkerschen:
There can be only one.

PR #343, opened 2019-02-11
BOAC-1795 Thread the needle on zero vs. null GPA

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1795


raydavis:
More like removing a needle from a haystack. Thanks, @paulkerschen

PR #344, opened 2019-02-12
NS-379 Improve the verification scheme for the generated data catalog

sandeepmjay:
The PR contains the following logic,

1) get an inventory of all S3 objects in a prefix for canvas-data.
2) Isolate unique table names from the s3 objects  extracted from the daily location
3) Run data verification processes to identify bad migration scenarios 

The logic ensures that all the canvas-data files sync are cataloged and verified. Tables for which schema exist but there is no data are omitted in the process.

https://jira.ets.berkeley.edu/jira/browse/NS-379

### DIFF
--- a/nessie/externals/redshift.py
+++ b/nessie/externals/redshift.py
@@ -95,13 +95,13 @@ def copy_tsv_from_s3(table, s3_key):
 
 
 def create_external_schema(external_schema, role):
-    query = """
+    query = f"""
         CREATE EXTERNAL SCHEMA {external_schema}
         FROM data catalog
         DATABASE '{external_schema}'
-        IAM_ROLE '{redshift_iam_role}'
+        IAM_ROLE '{role}'
         CREATE EXTERNAL DATABASE IF NOT EXISTS;
-        """.format(external_schema, role)
+        """
 
     if not execute(query):
         app.logger.error('Error on Redshift create schema')
--- a/nessie/jobs/refresh_canvas_data_catalog.py
+++ b/nessie/jobs/refresh_canvas_data_catalog.py
@@ -27,8 +27,8 @@ ENHANCEMENTS, OR MODIFICATIONS.
 from datetime import datetime, timedelta
 
 from flask import current_app as app
-from nessie.externals import canvas_data, redshift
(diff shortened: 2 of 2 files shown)


nessie/externals/redshift.py
```diff
@@ -96,12 +96,12 @@ def copy_tsv_from_s3(table, s3_key):
 
 def create_external_schema(external_schema, role):
     query = """
-        CREATE EXTERNAL SCHEMA {external_schema}
```
raydavis: Is there a particular reason to avoid our usual `f` quote here? Seems like it would be more readable.

nessie/jobs/refresh_canvas_data_catalog.py
```diff
@@ -139,3 +140,27 @@ def generate_external_catalog(self, external_schema, schema_df):
...
+        s3_client = s3.get_client()
+        bucket = app.config['LOCH_S3_BUCKET']
+        external_schema = app.config['REDSHIFT_SCHEMA_CANVAS']
+        prefix = get_s3_canvas_daily_path(datetime.now() - timedelta(days=1))
+        app.logger.info(f'Daily path = {prefix}')
+        directory_names = []
+        s3_objects = s3_client.list_objects_v2(Bucket=bucket, Prefix=prefix)
+        for object_summary in s3_objects['Contents']:
+            # parse table names from the S3 object URLs
+            directory_names.append(object_summary['Key'].split('/')[3:4][0])
```
paulkerschen: @sandeepmjay Any reason not to write this as `object_summary['Key'].split('/')[3]`?

nessie/externals/redshift.py
sandeepmjay: Switch to using f strings @raydavis.

nessie/jobs/refresh_canvas_data_catalog.py
sandeepmjay: Yep, makes sense. Looks like a hangover from my scratchpad. Switched it :)

sandeepmjay:
Amends are in @raydavis  @paulkerschen. Ready for merge.

PR #346, opened 2019-02-13
NS-380 Attempt to recover from S3 streaming failures during analytics calculation loop

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-380


sandeepmjay:
Looks good @paulkerschen. Shipping it.

PR #347, opened 2019-02-14
[QA] NS-380 Attempt to recover from S3 streaming failures during analytics calculation loop

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-380

QA version of #346.

I suspect this may be less needed with our larger instance size, but a run on dev doesn't suggest that it introduced any new problems.


raydavis:
Thanks, @paulkerschen

PR #348, opened 2019-02-16
NS-395 Make timestamp discrepancy tolerance a configured value

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-395

And expand tolerance further, because you can't trust the data to behave.


sandeepmjay:
Thanks @paulkerschen.

PR #351, opened 2019-02-19
[NS-396] Switching canvas schema creation with an automated schema generation job

sandeepmjay:
https://jira.ets.berkeley.edu/jira/browse/NS-396


johncrossman:
Thanks for talking it through, @sandeepmjay

PR #352, opened 2019-02-20
[NS-397] Added fail safes for canvas daily path generation

sandeepmjay:
https://jira.ets.berkeley.edu/jira/browse/NS-397

### DIFF
--- a/nessie/jobs/refresh_canvas_data_catalog.py
+++ b/nessie/jobs/refresh_canvas_data_catalog.py
@@ -97,13 +97,14 @@ class RefreshCanvasDataCatalog(BackgroundJob):
 
         if redshift.execute_ddl_script(canvas_external_catalog_ddl):
             app.logger.info(f'Canvas schema creation job completed.')
-            return self.verify_external_data_catalog()
+            if self.verify_external_data_catalog():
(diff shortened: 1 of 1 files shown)


nessie/jobs/refresh_canvas_data_catalog.py
```diff
@@ -97,13 +97,14 @@ def run(self):
 
         if redshift.execute_ddl_script(canvas_external_catalog_ddl):
             app.logger.info(f'Canvas schema creation job completed.')
-            return self.verify_external_data_catalog()
+            if self.verify_external_data_catalog():
+                return 'Canvas schema verfication job completed.'
```
paulkerschen: @sandeepmjay Do you need some code for the `else` condition here? (If `redshift.execute_ddl_script(canvas_external_catalog_ddl)` returns true but `self.verify_external_data_catalog()` returns false).

sandeepmjay: The verify external data catalog raises a background job error in case of verification failures.

paulkerschen: Thanks, looks good.

PR #357, opened 2019-02-23
BOAC-1858 Create full-text advising notes index in RDS

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1858

Meet dblink and to_tsvector! Postgres does it all.


johncrossman:
Postgres keeps on giving!

PR #359, opened 2019-02-28
NS-403 Add EDW advisee data to external schema

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-403

The job is not yet run on a regular schedule, and the data is not yet passed on in usable condition to BOA. But at least it can be made available to support further analysis and development.


paulkerschen:
So many schemas...

PR #360, opened 2019-02-28
NS-401: Adds future term enrollments to sis schema

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-401


paulkerschen:
The future is here!

PR #368, opened 2019-03-07
[QA] NS-435 Fix hardcoded schema name in template SQL

paulkerschen:
QA-first PR: https://jira.ets.berkeley.edu/jira/browse/NS-435


lyttam:
Thanks for catching.

PR #371, opened 2019-03-08
NS-429 Reorganize enrollment-term feed generation for scalability

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-429

Once again, we do all the same things in a different order.


raydavis:
Stand still faster.

PR #372, opened 2019-03-12
NS-431 NS-432 Split analytics generation/merge into self-contained per-term jobs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-432

I thought about breaking this up into multiple commits for review, but I think file-by-file is actually the easiest way to get through it.

The view from 30,000 feet:
- Our partition key of choice is term id, since that will allow for the most efficient reuse of data structures. Accordingly, many of our Redshift tables get an extra term column, large data structures are rearranged to sort by term first, and the giant S3-stored CSV of assignment submissions splits into several somewhat-giant term-specific CSVs.
- Enrollment term feed generation moves into its own GenerateMergedEnrollmentTerm job. For the moment each term is running on the master node, one after the next, but in a parallelized setup they'll be running simultaneously on worker nodes. To reduce contention on Redshift in this scenario, the master-node GenerateMergedStudentFeeds job queries for the required data structures in advance and writes them to S3 as JSON feeds for the worker nodes to pick up.
- The new GenerateMergedEnrollmentTerm job uses much of the same staged-upload-to-Redshift logic as the old GenerateMergedStudentFeeds job; common code is moved into nessie.models.student_schema.


paulkerschen:
Laggy integration but green Travis.

johncrossman:
Looks good to me @paulkerschen – a sensible structure. Thank you for detailed notes.

PR #373, opened 2019-03-12
NS-431 Master-to-worker job dispatch for analytics generation

paulkerschen:

Log management, queue tuning and all that will come in follow-ups as needed: this is just to see if the wiring works.


raydavis:
@paulkerschen , have we gotten data yet about the performance profile of your earlier refactor _without_ the extra worker instances?

paulkerschen:
@raydavis - Yes, see https://jira.ets.berkeley.edu/jira/browse/NS-429. We had an improvement of 2-3x specifically for the per-course analytics loop, which is about a 40% improvement for the job as a whole.

raydavis:
Thanks -- I wouldn't want to lock us into the extra AWS overhead unnecessarily, but 40% probably isn't quite enough on its own.

PR #374, opened 2019-03-12
NS-431 Master-to-worker job dispatch: how about using the right URL

paulkerschen:


raydavis:
Worth a shot.

PR #375, opened 2019-03-13
NS-443 Use cheaper query for Redshift status check

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-443


johncrossman:
We pay good money to have cheaper queries!

PR #376, opened 2019-03-15
NS-442 Database-backed job queue for enrollment terms merge

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-442

Same idea as the previous HTTP master-to-worker dispatch, but using shared Postgres should guarantee more even distribution and give us better reporting on results.


sandeepmjay:
Nice work @paulkerschen ! Especially with using RDS metadata switchover for enrollment terms. Shipping it !

PR #380, opened 2019-03-15
NS-449 Merged feeds job: fix final status formatting

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-449


johncrossman:
Final fixes fixed.

PR #381, opened 2019-03-18
NS-437 Remove superfluous vacuum commands

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-437

Since all these jobs run in sequence anyway, we really don't need to be so profligate with the vacuum cleaner.


johncrossman:
Time to throw out the vacuum bag.

PR #383, opened 2019-03-19
NS-452 Fix uncommitted transactions in enrollment terms refresh

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-452

### DIFF
--- a/nessie/models/student_schema.py
+++ b/nessie/models/student_schema.py
@@ -43,14 +43,12 @@ def staging_schema():
 
 
 def drop_staged_enrollment_term(term_id):
-    with redshift.transaction() as transaction:
-        delete_sql = 'DELETE FROM {schema}.{table} WHERE term_id = %s'
-        transaction.execute(
-            delete_sql,
-            schema=psycopg2.sql.Identifier(staging_schema()),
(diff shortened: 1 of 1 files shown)


nessie/models/student_schema.py
```diff
@@ -43,22 +43,18 @@ def staging_schema():
...
-            delete_sql,
-            schema=psycopg2.sql.Identifier(staging_schema()),
-            table=psycopg2.sql.Identifier('student_enrollment_terms'),
-            params=(term_id,),
-        )
+    redshift.execute(
+        'DELETE FROM {schema}.{table} WHERE term_id = %s',
+        schema=psycopg2.sql.Identifier(staging_schema()),
+        table=psycopg2.sql.Identifier('student_enrollment_terms'),
+        params=(term_id,),
```
johncrossman: @paulkerschen Is trailing comma, after `term_id`, necessary?

paulkerschen: @johncrossman Actually yes; it's a Python quirk with one-element tuples. https://ubuntuforums.org/archive/index.php/t-2232993.html

PR #384, opened 2019-03-20
NS-451 Multithread SIS Student API loop for scalability

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-451


sandeepmjay:
Very nice @raydavis.

PR #385, opened 2019-03-21
NS-455 Ensure analytics data structure for students without assignments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-455

Restore a bit of logic that got lost in the refactor.


johncrossman:
'Lost in a refactor', a Clash deep cut.

PR #387, opened 2019-03-22
NS-457, convert legacy note timestamps to proper UTC

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-457

We have 504,956 legacy notes. For now, this job is run as needed but suggestions on improving performance are welcome. Can _bulk_ RDS updates be made with current `psycopg`?


paulkerschen:
@johncrossman A bulk RDS update can be made with current psycopg, as long as we can figure out a way to express the transformation in pure SQL. But I'm wonder if we could make that transformation as part of the internal schema creation in create_sis_advising_notes_schema.template.sql. Seems like it would be best to correct the timestamp there, rather than having to update after the fact.

johncrossman:
@paulkerschen Yes, I was wondering the same. I'll investigate.

johncrossman:
I'll reopen when internal schema creation has the fix.

PR #388, opened 2019-03-25
NS-457, Redshift + Python UDF function: wonky date to UTC, with ISO format, during create schema

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-457

Python-based function in SQL template takes wonky Pacific time date, converts to UTC and returns in ISO format, and then finally, the built-in converts it to `TIMESTAMPTZ` type.

More info on Redshift + Python UDF:
https://docs.aws.amazon.com/redshift/latest/dg/udf-creating-a-scalar-udf.html

### DIFF
--- a/nessie/sql_templates/create_sis_advising_notes_schema.template.sql
+++ b/nessie/sql_templates/create_sis_advising_notes_schema.template.sql
@@ -277,6 +277,22 @@ CREATE SCHEMA {redshift_schema_sis_advising_notes_internal};
 -- Internal tables
 --------------------------------------------------------------------
 
+CREATE OR REPLACE FUNCTION {redshift_schema_sis_advising_notes_internal}.to_utc_iso_string(date_string VARCHAR)
+RETURNS VARCHAR
+STABLE
+AS $$
+  from datetime import datetime
+  import pytz
+
+  d = datetime.strptime(date_string, '%d-%b-%y %I.%M.%S.%f000 %p')
+  d = pytz.timezone('America/Los_Angeles').localize(d)
+  return d.astimezone(pytz.utc).isoformat()
+$$ language plpythonu;
+
+GRANT EXECUTE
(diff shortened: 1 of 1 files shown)


johncrossman:
I've been testing against `boac_advising_notes_johncrossman.advising_notes` so you can compare existing buggy data with what is generated by the change of this PR. 

A few sample notes with *existing code in dev*:
```
SELECT created_at, updated_at
FROM boac_advising_notes.advising_notes N
WHERE id LIKE '10627435-0000%'
ORDER BY id
```
 
The same notes converted with the code of this PR:
```
SELECT created_at, updated_at
FROM boac_advising_notes_johncrossman.advising_notes
WHERE id LIKE '10627435-0000%'
ORDER BY id
```

nessie/sql_templates/create_sis_advising_notes_schema.template.sql
```diff
@@ -277,6 +277,22 @@ CREATE SCHEMA {redshift_schema_sis_advising_notes_internal};
...
+  d = datetime.strptime(date_string, '%d-%b-%y %I.%M.%S.%f000 %p')
+  d = pytz.timezone('America/Los_Angeles').localize(d)
+  return d.astimezone(pytz.utc).isoformat()
+$$ language plpythonu;
+
+GRANT EXECUTE
+ON function {redshift_schema_sis_advising_notes_internal}.to_utc_iso_string(VARCHAR)
+TO GROUP {redshift_app_boa_user}_group;
+
 CREATE TABLE {redshift_schema_sis_advising_notes_internal}.advising_notes
```
johncrossman: The use of "STABLE" above is explained at https://docs.aws.amazon.com/redshift/latest/dg/r_CREATE_FUNCTION.html

paulkerschen:
Seems like pytz is doing the right thing summer and winter. Thank you @johncrossman!

PR #389, opened 2019-03-25
NS-460 Don't unmerge assignment submissions after we worked so hard to merge them

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-460

#385 was missing a rather important line.


sandeepmjay:
Nessie is an eager beaver. Thanks @paulkerschen !

PR #394, opened 2019-04-05
BOAC-1917, expand LDAP/CalNet query to include 'expired people'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1917

See description of https://github.com/ets-berkeley-edu/boac/pull/1412


raydavis:
What's better than an expired population? An even LARGER expired population!

PR #395, opened 2019-04-08
NS-477 Fix exception in LDAP processing

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-477


johncrossman:
Thank you for the fix @raydavis

PR #396, opened 2019-04-11
NS-445 Move metadata schema from Redshift to RDS

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-445


scripts/migrate/20190410-NS-445/copy_rds_metadata_from_redshift.sql
```diff
@@ -0,0 +1,125 @@
+-- This script needs the "redshift_db" variable set before being run.
```
raydavis: This seemed a bit more in line with our release conventions than writing a shell script, but let me know if you disagree.

nessie/jobs/create_metadata_schema.py
```diff
@@ -44,11 +44,4 @@ def run(self):
             raise BackgroundJobError(f'RDS metadata schema creation failed.')
             return False
 
-        resolved_ddl_redshift = resolve_sql_template('create_metadata_schema_redshift.template.sql')
```
johncrossman: @raydavis - Is `create_metadata_schema_redshift.template.sql` still used?

scripts/migrate/20190410-NS-445/copy_rds_metadata_from_redshift.sql
johncrossman: I agree.

nessie/jobs/create_metadata_schema.py
raydavis: Not by code, @johncrossman , so I should delete it, even if the Redshift config variables stay around for a release. I'll amend.

tests/conftest.py
```diff
@@ -107,9 +108,12 @@ def metadata_db(app):
         created_at TIMESTAMP NOT NULL,
         deleted_at TIMESTAMP
     )""")
-    rds_schema = app.config['RDS_SCHEMA_METADATA']
-    rds.execute(f'DROP SCHEMA IF EXISTS {rds_schema} CASCADE')
-    rds.execute(f'CREATE SCHEMA IF NOT EXISTS {rds_schema}')
+    rds.execute(f"""CREATE TABLE IF NOT EXISTS {rds_schema}.merged_feed_status
+    (
+        sid VARCHAR NOT NULL PRIMARY KEY ,
```
johncrossman: @raydavis Minor: remove space before comma

johncrossman:
@raydavis Looks good. Can we remove `REDSHIFT_SCHEMA_METADATA` from `config/*.py` files?

raydavis:
@johncrossman , I had planned to keep the REDSHIFT_SCHEMA_METADATA configs in place so as to support copying old Redshift data to the new RDS tables. However, since we're going to try doing the copy with plain SQL (plus one variable), I think we can remove them after all.

I don't yet know what's up with Travis, since local runs work for me. But I'll keep poking at it for a bit.

raydavis:
@johncrossman Looks like that last amendment pleased Travis.

raydavis:
... but we'll hold off on a merge until tomorrow morning, to avoid conflicts with nightly jobs.

PR #397, opened 2019-04-15
NS-492 Do not stop nightly jobs when faced with a new incoming student

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-492


johncrossman:
New kids on the block ain't _that_ special.

PR #400, opened 2019-04-25
NS-499: Adds schemas for ASC advising notes

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-499


paulkerschen:
Looks like big data to me....

PR #412, opened 2019-05-20
NS-540: Points to new S3 location for historical SIS notes

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-540

The historical snapshot of SIS notes data has been copied to s3://la-nessie-protected-dev/sis-data/historical/advising-notes.  This PR updates the code to point to the new location.  

I also updated the nessie-dev deploy config accordingly.  Once deployed and tested, I'll remove the old S3 folders that will no longer be in use.

### DIFF
--- a/nessie/lib/util.py
+++ b/nessie/lib/util.py
@@ -103,7 +103,7 @@ def get_s3_sis_api_daily_path(cutoff=None):
 def resolve_sql_template_string(template_string, **kwargs):
     """Our DDL template files are simple enough to use standard Python string formatting."""
     s3_prefix = 's3://' + app.config['LOCH_S3_BUCKET'] + '/'
-    s3_advising_notes_prefix = 's3://' + app.config['LOCH_S3_SIS_ADVISING_NOTES_BUCKET'] + '/'
+    s3_protected_prefix = 's3://' + app.config['LOCH_S3_PROTECTED_BUCKET'] + '/'
     template_data = {
         'rds_app_boa_user': app.config['RDS_APP_BOA_USER'],
         'rds_dblink_to_redshift': app.config['REDSHIFT_DATABASE'] + '_redshift',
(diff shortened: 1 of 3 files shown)


nessie/lib/util.py
```diff
@@ -146,9 +146,9 @@ def resolve_sql_template_string(template_string, **kwargs):
         'loch_s3_canvas_data_path_current_term': s3_prefix + app.config['LOCH_S3_CANVAS_DATA_PATH_CURRENT_TERM'],
         'loch_s3_coe_data_path': s3_prefix + get_s3_coe_daily_path(),
         'loch_s3_physics_data_path': s3_prefix + app.config['LOCH_S3_PHYSICS_DATA_PATH'],
-        'loch_s3_sis_advising_notes_data_path': s3_advising_notes_prefix + app.config['LOCH_S3_SIS_ADVISING_NOTES_DATA_PATH'],
         'loch_s3_sis_data_path': s3_prefix + app.config['LOCH_S3_SIS_DATA_PATH'],
         'loch_s3_sis_api_data_path': s3_prefix + get_s3_sis_api_daily_path(),
+        'loch_s3_sis_data_protected_path': s3_advising_notes_prefix + app.config['LOCH_S3_SIS_DATA_PATH'],
```
paulkerschen: @lyttam Should we perhaps change the variable name `s3_advising_notes_prefix` (and the config value `LOCH_S3_SIS_ADVISING_NOTES_BUCKET`) to reflect their new function as all protected, all the time?

lyttam: @paulkerschen I thought about that too, though it would make more work for Darlene. But I suppose we'll want to change it eventually.

lyttam: @paulkerschen amended with your suggestion. I'll update nessie-dev.py as well.

paulkerschen:
Thanks, @lyttam, looks good.

PR #423, opened 2019-05-31
NS-547: Wraps SIS note indexing in transaction

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-547


paulkerschen:
Thanks for extra code readability improvements, @lyttam.

PR #431, opened 2019-06-04
BOAC-2298 Fix floating point cruft when summing units

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2298


paulkerschen:
As many float accidents as the Macy's parade.

PR #433, opened 2019-06-04
[QA] NS-555 Rerun IndexEnrollments after student population update

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-555


raydavis:
The new Delete Job endpoint finds a use!

PR #434, opened 2019-06-04
[QA] BOAC-2298 Fix floating point cruft when summing units

raydavis:
QA version of https://github.com/ets-berkeley-edu/nessie/pull/431


johncrossman:
If the cruft does not bring joy then out it goes.

PR #435, opened 2019-06-05
BOAC-2238: Adds advising note search index to ASC schema

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2238

I thought it made sense to add these tables to the existing `boac_advising_asc` schema, but I could alternatively add them to the `boac_advising_notes` schema or create a new schema for them.


paulkerschen:
Putting them in the ASC schema works for me; after all, who knows what kind of data we'll be ingesting from these various partners in the future. Perhaps baseball statistics are next.

PR #439, opened 2019-06-11
NS-562 Restore future-terms processing to GenerateMergedStudentFeeds

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-562


lyttam:
Thanks for the investigation and restoration!

PR #447, opened 2019-06-14
NS-570 You gotta truncate em afore you load em

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-570


johncrossman:
And know when to fold 'em...

PR #451, opened 2019-06-20
BOAC-1081: Welcomes legacy SIS enrollments into the loch

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1081


nessie/jobs/generate_merged_student_feeds.py
```diff
@@ -97,14 +97,14 @@ def generate_feeds(self):
         feed_path = app.config['LOCH_S3_BOAC_ANALYTICS_DATA_PATH'] + '/feeds/'
         s3.upload_json(advisees_by_canvas_id, feed_path + 'advisees_by_canvas_id.json')
 
-        # Avoid processing Canvas analytics data for future terms and pre-CS terms.
-        for term_id in future_term_ids():
+        term_ids = reverse_term_ids(include_future_terms=True, include_legacy_terms=True)
+        for term_id in term_ids:
             GenerateMergedEnrollmentTerm().refresh_student_enrollment_term(term_id, enrollment_terms_map.get(term_id, {}))
+            s3.upload_json(enrollment_terms_map.get(term_id, {}), feed_path + f'enrollment_term_map_{term_id}.json')
```
paulkerschen: @lyttam This is tricky, but there might be some redundant work going on here. I think we need to explicitly call `GenerateMergedEnrollmentTerm().refresh_student_enrollment_term` for future terms and legacy terms, but not for those terms included in our usual `reverse_term_ids()`, because those terms will get refreshed with analytics as a consequence of the `queue_merged_enrollment_term_jobs` call below.

Conversely, the `s3.upload_json(enrollment_terms_map...` only needs to happen for our usual `reverse_term_ids()`. We stash that data in S3 so that worker nodes can pick it up, and it's only our expensive analytics-containing terms that are distributed among worker nodes by the `queue_merged_enrollment_term_jobs` logic.

lyttam: Thanks @paulkerschen, I wasn't quite sure about that bit. Will amend.

raydavis:
[This line](https://github.com/ets-berkeley-edu/nessie/blob/3f55645af7959b81a8a6580bf3b1930fbd4debab/nessie/jobs/generate_merged_student_feeds.py#L95) of `generate_merged_student_feeds.py` probably needs rethinking to avoid the call to `get_canvas_site_maps()` for pre-CS terms. For future terms, it does no harm (even if we're not using the sites to provide analytics), but for past terms the bad section-ID mapping sets us up for some confusing bugs.

paulkerschen:
@raydavis I haven’t stepped through the code all the way, but it looks like we get our initial Canvas site info from nessie.lib.queries.get_advisee_enrolled_canvas_sites, which seems to exclude future and pre-CS terms.

raydavis:
By gum, you're right, @paulkerschen . Sorry about the false alarm, @lyttam .

lyttam:
@raydavis No problem, I just revisited that query to double check.

nessie/lib/berkeley.py
```diff
@@ -162,14 +162,11 @@ def canvas_terms():
...
-    while True:
-        if term_id <= stop_term_id:
-            return term_ids
-        term_ids.append(term_id)
-        term_id = previous_term_id(term_id)
+    return _collect_terms(start_term_id=future_term_id(), stop_term_id=current_term_id(), inclusive=False)
+
+
+def legacy_term_ids():
+    return _collect_terms(start_term_id=earliest_term_id(), stop_term_id=earliest_legacy_term_id())
```
paulkerschen: @lyttam Sorry for the delayed review - it looks like there might be an inclusive gotcha at the start as well as the end of the loop? I think this code would count Fall 2016 as both a legacy and a non-legacy term.

paulkerschen:
Apart from one question about legacy term overlap, all looks good to me, @lyttam - thanks!

nessie/lib/berkeley.py
lyttam: @paulkerschen I guess we don't want that.

lyttam: @paulkerschen amended to fix that overlap.

PR #456, opened 2019-06-24
NS-583 Preserve associated SIS enrollments when filtering out unpublished Canvas sites

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-583


johncrossman:
Ooo, tricky.

PR #457, opened 2019-06-24
NS-580: Discards previous versions of an ASC note

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-580

When a note is edited after it is created, ASC will send us two records - we want only the most recent one.


paulkerschen:
Versioning!

PR #461, opened 2019-06-25
NS-584 Retry S3 read after a timeout error

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-584

Because "S3 will do that occasionally."


paulkerschen:
Why not!

PR #462, opened 2019-06-27
NOJIRA: Fixes table name mishap

lyttam:
I happened upon what looks like just a copy-paste error.


paulkerschen:
Thanks, @lyttam - that would have been an awfully annoying bug to track down if we'd hit unlucky timing.

PR #466, opened 2019-07-01
NS-553 Add non-CoE-specific Gender, Ethnicity data to Loch

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-553

(Junction-Dev has already uploaded the necessary `sis-data/advisees/demographics` data to Nessie-Dev's S3.)


paulkerschen:
What a great lot of demography!

PR #468, opened 2019-07-08
NS-575: Adds SIS note import to schedule

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-575

After Junction j104 is released, SIS notes data will be uploaded to S3 at 8pm each day.  I think it makes sense to group all the Nessie notes jobs together, so the SIS notes schema will be created right after the ASC notes schema at 3:15am.


paulkerschen:
@lyttam Looks like you got hit with the same pylint surprise that I did this morning. See https://github.com/ets-berkeley-edu/boac/commit/be837490fdc6b79c3381a84e95f2726c35d9b7ea

lyttam:
Darn. "Works on my machine".

PR #470, opened 2019-07-09
BOAC-2352: Creates internal schema for advisor data

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2352

This is just a first pass and will likely need refinement as BOA requirements become clearer.


paulkerschen:
Thanks for the start, @lyttam.

PR #472, opened 2019-07-11
BOAC-2351: Adds advisor dept to internal schema

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2351

Since the story that this Jira belongs to got moved to the next sprint, I'm not planning to submit a QA PR for this.


paulkerschen:
A cad, a plan, a canal.

PR #476, opened 2019-07-15
NS-478 Switch to V2 SIS Students collections API

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-478 (and its subtasks)

* Replaces current student profile loop with V2 "list" APIs.
* Change Terms GPA job to store most recent registration data as well, since that's needed to fully replace the V1 "currentRegistrations" element.
* Backfills are now handled by a regularly scheduled ImportRegistrations job rather than being shoehorned into GenerateMergedStudentFeeds.

The sis_student_api module includes a couple of not-used-by-application-code functions to help with diagnostics.


paulkerschen:
Glad we're logging all those warnings! Thanks, @raydavis, let's see how it flies.

PR #483, opened 2019-07-22
NOJIRA, Nessie is now v3.0 on master branch

johncrossman:
With Fall 2019 comes Nessie/BOA v3.0 and an aligned versioning scheme.


lyttam:
A major milestone!

PR #484, opened 2019-07-22
BOAC-2507: Updates advisor_roles logic

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2507

### DIFF
--- a/nessie/sql_templates/create_advisor_schema.template.sql
+++ b/nessie/sql_templates/create_advisor_schema.template.sql
@@ -165,7 +165,7 @@ AS (
 );
 
 CREATE TABLE {redshift_schema_advisor_internal}.advisor_roles
-SORTKEY (sid, uid)
+SORTKEY (sid)
 AS (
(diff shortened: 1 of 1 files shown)


nessie/sql_templates/create_advisor_schema.template.sql
```diff
@@ -174,10 +174,11 @@ AS (
         I.ADVISOR_TYPE_DESCR AS advisor_type,
         I.INSTRUCTOR_TYPE AS instructor_type_code,
         I.INSTRUCTOR_TYPE_DESCR AS instructor_type,
-        P.PERMISSION_LIST AS cs_permissions,
-        P.AUTH_ACTIONS AS cs_auth_actions
```
lyttam: Removed AUTH_ACTIONS because I didn't include that field in the file in S3 (since the value appears to be meaningless).

paulkerschen:
Thanks @lyttam!

PR #485, opened 2019-07-23
[NS-534]  Use boto sessions for S3 operations. Smart open update to 1.8.3.

sandeepmjay:
Update smart open to 1.8.3
Use sts to create boto sessions for S3 authentication
Update tests to use mock_sts
update config.py to include APP_ROLE
Parameterize Encryption on S3 to use AES256

https://jira.ets.berkeley.edu/jira/browse/NS-534

### DIFF
--- a/nessie/externals/s3.py
+++ b/nessie/externals/s3.py
@@ -38,13 +38,9 @@ import smart_open
 """Client code to run file operations against S3."""
 
 
-def build_s3_url(key, credentials=True):
+def build_s3_url(key):
     bucket = app.config['LOCH_S3_BUCKET']
-    if credentials:
-        credentials = ':'.join([app.config['AWS_ACCESS_KEY_ID'], app.config['AWS_SECRET_ACCESS_KEY']])
-        return f's3://{credentials}@{bucket}/{key}'
-    else:
-        return f's3://{bucket}/{key}'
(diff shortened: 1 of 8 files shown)


nessie/externals/s3.py
```diff
@@ -204,14 +216,20 @@ def upload_from_url(url, s3_key, on_stream_opened=None):
...
                     'ContentEncoding': 'gzip',
                     'ContentType': 'text/plain',
                 })
+            session = get_session()
             # smart_open needs to be told to ignore the .gz extension, or it will smartly attempt to double-compress it.
-            with smart_open.smart_open(s3_url, 'wb', ignore_extension=True, s3_upload_args=s3_upload_args) as s3_out:
+            with smart_open.open(
+                s3_url,
+                'wb',
+                ignore_ext=True,
```
johncrossman: @sandeepmjay `ignore_ext` is right and `ignore_extension` is wrong?

johncrossman:
Looks good @sandeepmjay  I put one question.

nessie/externals/s3.py
sandeepmjay: Smart open folks thought it would optimize to maximize the smartness with the change in 1.8.3

johncrossman: Smart they are.

PR #486, opened 2019-07-23
[NS-550] Parameterized Elastic Beanstalk instance profiles. 

sandeepmjay:
parameterizes IAM Instance profile supplied to EC2 instances
Set default region for S3 smart open interactions
Install AWS Inspector agent to track and monitor EC2 instances for resource usage and security monitoring.

https://jira.ets.berkeley.edu/jira/browse/NS-550


paulkerschen:
Let’s give it a go.

PR #487, opened 2019-07-23
BOAC-2505: Adds advisor schema job to schedule

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2505


paulkerschen:
13 lucky jobs.

PR #492, opened 2019-07-25
[NS-621] Downgrade boto log propagation level as debug mode is extra verbose

sandeepmjay:
https://jira.ets.berkeley.edu/jira/browse/NS-621


raydavis:
Thanks, @sandeepmjay

PR #495, opened 2019-07-26
BOAC-2551: Adds column to differentiate between historical and incremental SIS attachments

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2551


paulkerschen:
Of historical importance.

PR #499, opened 2019-08-02
NS-635 Fix withdrawalCancel in SIS Profile feed

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-635


paulkerschen:
Personal, not persnickety.

PR #501, opened 2019-08-02
BOAC-2393: Imports E&I notes

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2393


nessie/jobs/index_advising_note_authors.py
```diff
@@ -0,0 +1,88 @@
...
+        asc_schema = app.config['RDS_SCHEMA_ASC']
+        e_i_schema = app.config['RDS_SCHEMA_E_I']
+        sis_schema = app.config['RDS_SCHEMA_SIS_ADVISING_NOTES']
+
+        with rds.transaction() as transaction:
+            if not transaction.execute(f'TRUNCATE {sis_schema}.advising_note_author_names'):
+                transaction.rollback()
+                raise BackgroundJobError('Failed to truncate advising note author name index.')
+
+            sql = f"""INSERT INTO {sis_schema}.advising_note_author_names (
```
lyttam: Should the `advising_note_author_names` table be alone in its own schema, or does it make sense to keep it in `boac_advising_notes`?  Everything else in that schema is SIS-specific.

paulkerschen: @lyttam Per my TODO in the old code, I stuck that table in the SIS-specific schema because I was being lazy at the time, but I'd say it's grown up enough to deserve its own non-dept-specific advisor schema.

lyttam: Thanks for the validation. Will amend on Monday.

lyttam:
@paulkerschen finally got around to amending this; now there's a second commit. Do you think it's a good idea to also add a migration script to drop the existing notes schemas so we can start fresh when the jobs run?

paulkerschen:
Thanks for this, @lyttam. I agree that we definitely want to document any required schema dropping; but if it's just a matter of removing a few schemas before the first job run, it might suffice to note those drops in release instructions. If we have to do anything more complex, a follow-up migration script might be warranted.

PR #502, opened 2019-08-05
NS-637 Fall back to V1 SIS Students API

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-637

Because the V2 API integration tried to keep upper-layer data feeds stable, the V1 reversion can (theoretically) be swapped back in without much disruption. As usual with this massive operation, though, the only way to test that assumption is to try it on a running system and live with any delays that result.

In other words, a close code review would be even more appreciated than usual.


paulkerschen:
I detect no problems with the naked eye. Thanks for the extra work to feature-flag this; I guess we'd better see how it runs.

PR #504, opened 2019-08-05
NS-637 Avoid Law careers, since they currently break upstream logic

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-637


paulkerschen:
Good career advice.

PR #508, opened 2019-08-08
BOAC-2393: Moves DROP TABLE commands to release instructions

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2393

These commands need to be run once, so it makes more sense to do them as a release step.


paulkerschen:
Makes sense.

PR #509, opened 2019-08-08
BOAC-2395: Expands advising_note_authors to include ASC and E&I

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2395

This moves the advising_note_authors table from the SIS-specific schema to the general advising notes schema and adds advisors from ASC and E&I.  This will make it easier to return ASC and E&I notes in a search by author in BOA.


paulkerschen:
Thanks for freeing these names from their silos, @lyttam.

PR #510, opened 2019-08-08
BOAC-2396: Adds E&I notes creation and note author indexing to schedule

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2396


paulkerschen:
Thanks, @lyttam - as you're no doubt aware, this will require a release instruction to update JOB_LOAD_ADVISING_NOTES (which can now be done from the schedule tab on Nessie's web console - first "Remove," then "Reload from configs.")

lyttam:
Thanks @paulkerschen , I needed the reminder.

PR #513, opened 2019-08-12
NS-642 Deal with incoming transfer units in V2 SIS Students API

raydavis:
See https://jira.ets.berkeley.edu/jira/browse/NS-642 _if you dare..._


lyttam:
I cannot un-see.

PR #514, opened 2019-08-13
NS-589: Refreshes RDS from Redshift rather than memory

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-589

This is part 1 of the work.


raydavis:
Thanks, @lyttam -- let's see how it runs tonight!

PR #515, opened 2019-08-15
NS-589: Avoids cluttering memory with excess enrollment terms

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-589

If this approach works well, we can do something similar with the profile feeds in `generate_student_profile_tables`.


paulkerschen:
Let's give it a go!

PR #516, opened 2019-08-16
NS-631: Begins stashing relevant enrollment terms in S3 for future analysis

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-631


paulkerschen:
Thanks, @lyttam, looks good to me. Does @sandeepmjay want to take a look?

sandeepmjay:
Looks good @lyttam on the code level. However, the S3 organization in the config we may need to make a change. Not a blocker for merge at this time. Thanks for the work!

PR #520, opened 2019-08-19
BOAC-2604, fix Y2K bug in Nessie

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2604

See Jira above for UID of student with expected grad term < 2000.


paulkerschen:
image

PR #523, opened 2019-08-22
BOAC-2623 RDS copies of BOA-accessed Redshift tables

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2623


raydavis:
We have SQL templates the way some places have mice.

PR #526, opened 2019-08-23
NS-663 Log cleaner saner SQL

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-663

### DIFF
--- a/nessie/externals/rds.py
+++ b/nessie/externals/rds.py
@@ -117,4 +117,4 @@ def _log_db_error(e, sql):
     if e.pgcode:
         error_str += f'{e.pgcode}: {e.pgerror}\n'
     error_str += f'on SQL: {sql}'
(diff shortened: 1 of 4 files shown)


nessie/externals/rds.py
```diff
@@ -117,4 +117,4 @@ def _log_db_error(e, sql):
     if e.pgcode:
         error_str += f'{e.pgcode}: {e.pgerror}\n'
     error_str += f'on SQL: {sql}'
-    app.logger.warn({'message': error_str})
+    app.logger.warning(error_str)
```
paulkerschen: I don't know why we were shoving our SQL errors into JSON objects; it must have been thoughtlessly carried over from Node, and sure does make the logs harder to read.

raydavis:
And you saved the Regents' investment! Look for a Spot Award in your future.

PR #536, opened 2019-08-28
NS-658 Consolidate loch-side advising notes search index

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-658


johncrossman:
You had me at "consolidate".

PR #537, opened 2019-08-28
NS-672: Creates job to copy attachments from incremental to historical

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-672


paulkerschen:
Historical attachments are written by the victors.

PR #540, opened 2019-08-29
NS-679 Fix schema interpolation in SQL template

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-679


raydavis:
A sadly familiar mistake.

PR #545, opened 2019-09-03
NS-669 Guard against massively swollen Student API feeds

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-669

Icky.


paulkerschen:
Sink those subs!

PR #546, opened 2019-09-03
NS-671: Includes non-current students on request (from BOA)

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-671

### DIFF
--- a/tests/test_jobs/test_import_student_photos.py
+++ b/tests/test_jobs/test_import_student_photos.py
@@ -37,7 +37,7 @@ class TestImportStudentPhotos:
         with capture_app_logs(app):
             with mock_s3(app):
                 result = ImportStudentPhotos().run_wrapped()
-                assert result == 'Student photo import completed: 1 succeeded, 8 had no photo available, 0 failed.'
+                assert result == 'Student photo import completed: 1 succeeded, 9 had no photo available, 0 failed.'
                 response = s3.get_keys_with_prefix('cal1card-data/photos')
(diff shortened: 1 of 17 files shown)


tests/test_jobs/test_import_student_photos.py
```diff
@@ -47,7 +47,7 @@ def test_import_student_photos(self, app, metadata_db, student_tables, caplog):
             assert success_rows[0]['sid'] == '11667051'
 
             failure_rows = rds.fetch(f"SELECT * FROM {app.config['RDS_SCHEMA_METADATA']}.photo_import_status WHERE status = 'failure'")
-            assert len(failure_rows) == 0
+            assert len(failure_rows) == 1
```
paulkerschen: @lyttam - Looking back at the code for ImportStudentPhotos, it seems like it's reporting a failure (rather than simply returning no response) because it can't find an UID translation for this SID, which it does by checking the Calnet schema. Should the new user be mocked out in that schema `(in fixtures/students.sql)` as well?

paulkerschen:
Thanks, @lyttam - this looks really good. Left one question about test data.

tests/test_jobs/test_import_student_photos.py
lyttam: @paulkerschen yes, good call. I'll amend.

lyttam:
Amended, @paulkerschen. Thanks!

PR #548, opened 2019-09-04
 NS-670 Generate BOA-friendly feeds for non-advisee student data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-670

### DIFF
--- a/nessie/merged/sis_profile.py
+++ b/nessie/merged/sis_profile.py
@@ -69,18 +69,18 @@ def merge_sis_profile_academic_status(sis_student_api_feed, sis_profile):
     # The Hub may return multiple academic statuses. We'll prefer an undergraduate enrollment, but
     # otherwise select the first well-formed status that is not a Law enrollment.
     academic_status = None
-    career_code = None
     for status in sis_student_api_feed.get('academicStatuses', []):
-        career_code = status.get('studentCareer', {}).get('academicCareer', {}).get('code')
-        if career_code and career_code == 'UGRD':
+        status_code = status.get('studentCareer', {}).get('academicCareer', {}).get('code')
+        if status_code and status_code == 'UGRD':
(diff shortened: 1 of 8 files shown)


nessie/merged/sis_profile.py
```diff
@@ -69,18 +69,18 @@ def merge_sis_profile_academic_status(sis_student_api_feed, sis_profile):
     # The Hub may return multiple academic statuses. We'll prefer an undergraduate enrollment, but
     # otherwise select the first well-formed status that is not a Law enrollment.
     academic_status = None
-    career_code = None
```
raydavis: The old code kept Law enrollments out of `academic_status` but let "LAW" slip through as the `career_code`.

raydavis:
Just amended this to fill in first and last name fields for BOA use.

nessie/models/student_schema.py
```diff
@@ -64,7 +64,7 @@ def refresh_from_staging(table, term_id, sids, transaction, truncate_staging=Tru
     # matching those restrictions. If there are no restrictions, the entire destination table can be truncated.
     refresh_conditions = []
     refresh_params = []
-    if (term_id and table == 'student_enrollment_terms'):
+    if term_id:
```
lyttam: @raydavis Will removing the `table == 'student_enrollment_terms'` cause a problem when this method is called for the student profile tables?  In `GenerateMergedStudentFeeds.generate_feeds`, we call `student_schema.refresh_all_from_staging(profile_tables)` where `profile_tables` includes student_profiles, student_academic_status, etc. - tables that don't have a `term_id` column.

lyttam: I think I just answered my own question. `term_id` will be None in those cases, so the `term_id` condition won't get appended.

PR #552, opened 2019-09-05
NS-672: Fixes timezone issue and accounts for maintenance days

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-672


paulkerschen:
Thanks for the test code, which helps make all this comprehensible.

PR #556, opened 2019-09-06
NS-685 Teach Degree-parsing code about the pre-CS world

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-685


raydavis:
Another hurdle before BOA can see all those aged ex-students, @paulkerschen .

paulkerschen:
I'd like to teach the world to parse.

PR #558, opened 2019-09-08
NS-670 Fix running count of non-advisee profiles

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-670


nessie/jobs/generate_merged_hist_enr_feeds.py
```diff
@@ -90,8 +90,9 @@ def generate_student_profile_table(self, unmerged_sids):
             if profile_count:
                 student_schema.truncate_staging_table(table_name)
                 student_schema.write_file_to_staging(table_name, feed_file, profile_count)
-        student_schema.refresh_all_from_staging([table_name])
+        profile_count = self.refresh_all_from_staging([table_name])
```
paulkerschen: @raydavis Should this be called on `student_schema` rather than `self`?

raydavis: Ugh, what a dopey mistake. Thank you for catching that, @paulkerschen .

PR #566, opened 2019-09-13
[QA] NS-695: Handles error from BOA API

lyttam:
QA version of #565 

https://jira.ets.berkeley.edu/jira/browse/NS-695


raydavis:
D'oh!

PR #574, opened 2019-09-18
BOAC-2753, until SISRP-48560 is resolved, we treat 0.5 as 12, the expected min-units per term

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2753

I changed keys `unitsMaxOverride` and `unitsMinOverride` to  `unitsMax` and `unitsMin` because they are not necessarily "overrides" – the attributes might carry default values. If this PR is merged then I'll submit (1) a QA PR and (2) the necessary BOA PRs.


paulkerschen:
Thanks for the comment clearly explaining as much of the mystery as we understand.

PR #576, opened 2019-09-19
NS-702 None cannot float

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-702

Nothingness is a heavy responsibility.


paulkerschen:
It's a witch!

PR #579, opened 2019-09-23
NS-707 Non-advisee list should include SIDs with only dropped classes

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-707


paulkerschen:
No enrollee left behind.

PR #582, opened 2019-09-25
NS-696: Standardizes time zones across SIS attachment jobs

lyttam:

The Migrate and Verify jobs now share the code that constructs a range of dates from the last successful run of the Migrate job to today.

The dates are converted from UTC to PST to match the dated folder structure created by Darlene's script, which runs in PST.


paulkerschen:
Standardization is good. Thanks for the test coverage, @lyttam.

PR #583, opened 2019-09-26
NS-696: Further simplifies the verify attachments job

lyttam:

Whereas the "since last successful run" logic has proved problematic for the VerifySisAdvisingNoteAttachments job, and whereas the work it does is inexpensive (relative to the Migrate job's work of copying files), the job will now validate all SIS attachments in S3 by default. 
 We can still narrow it down to a specific year, month, or date by running it manually with a datestamp param.


paulkerschen:
It is so ordered.

PR #585, opened 2019-09-27
NS-696: Ignores time when determining dates for attachment sync

lyttam:
Fun with datetimes.


paulkerschen:
Date, time, repeat.

PR #590, opened 2019-10-07
NS-682 Handle double-digit term GPAs when provided by SIS

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-682

You can't fight the Registrar.


paulkerschen:
Whatever the boss says.

PR #596, opened 2019-10-15
NS-719 Communicate an erroneously created affiliation more clearly

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-719


paulkerschen:
God don't make no junk, but he don't always get affiliations right either.

PR #598, opened 2019-10-23
NS-690 Automate term configurations in Nessie

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-606


paulkerschen:
Thanks much, @raydavis. It does seem to make sense.

PR #599, opened 2019-10-24
NS-690 Eliminate remaining dependencies on static term configurations

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-690


johncrossman:
Makes sense to me. Thanks @raydavis

PR #600, opened 2019-10-24
[NS-712] Canvas historical requests sync job for archival purposes.

sandeepmjay:
https://jira.ets.berkeley.edu/jira/browse/NS-712

### DIFF
--- a/nessie/jobs/sync_canvas_requests_snapshots.py
+++ b/nessie/jobs/sync_canvas_requests_snapshots.py
@@ -23,7 +23,6 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-import re
 import time
 
 from flask import current_app as app
@@ -31,16 +30,15 @@ from nessie.externals import canvas_data, s3
 from nessie.jobs.background_job import BackgroundJob, BackgroundJobError
 from nessie.lib import metadata
 from nessie.lib.dispatcher import dispatch
-from nessie.lib.util import get_s3_canvas_daily_path
 
 """Logic for Canvas snapshot sync job."""
 
 
-class SyncCanvasSnapshots(BackgroundJob):
+class SyncCanvasRequestsSnapshots(BackgroundJob):
 
     @classmethod
     def generate_job_id(cls):
-        return 'sync_' + str(int(time.time()))
+        return 'sync_canvas_requests_' + str(int(time.time()))
 
     def run(self, cleanup=True):
         job_id = self.generate_job_id()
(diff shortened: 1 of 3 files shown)


nessie/jobs/sync_canvas_requests_snapshots.py
```diff
@@ -0,0 +1,111 @@
...
+    app.logger.info(
+        f'Found {len(existing_keys)} key(s) matching prefix "{prefix}", {len(existing_keys) - len(keys_to_delete)} '
+        f'key(s) in whitelist, will delete {len(keys_to_delete)} object(s)')
+    if not keys_to_delete:
+        return True
+    if s3.delete_objects(keys_to_delete):
+        metadata.delete_canvas_snapshots(keys_to_delete)
+        return True
+    else:
+        return False
```
paulkerschen: @sandeepmjay It looks like this method is the same as the identically named method in the `SyncCanvasSnapshots` job. If this job in fact needs this method, I'd suggest moving it into a shared location.

nessie/jobs/sync_canvas_requests_snapshots.py
```diff
@@ -0,0 +1,111 @@
...
+        failure = 0
+
+        for snapshot in snapshots_to_sync:
+            metadata.create_canvas_sync_status(
+                job_id=job_id,
+                filename=snapshot['filename'],
+                canvas_table=snapshot['table'],
+                source_url=snapshot['url'],
+            )
+            if snapshot['table'] == 'requests':
```
paulkerschen: @sandeepmjay This conditional shouldn't be necessary if `snapshot['table']` will always be `requests`?

nessie/jobs/sync_canvas_requests_snapshots.py
```diff
@@ -0,0 +1,111 @@
...
+
+            if not response:
+                app.logger.error('Failed to dispatch S3 sync of snapshot ' + snapshot['filename'])
+                metadata.update_canvas_sync_status(job_id, key, 'error', details=f'Failed to dispatch: {response}')
+                failure += 1
+            else:
+                app.logger.info('Dispatched S3 sync of snapshot ' + snapshot['filename'])
+                success += 1
+
+        if cleanup:
```
paulkerschen: @sandeepmjay If the job is for archival purposes, will it need to make use of this cleanup routine?

sandeepmjay: It might if canvas changes the list of files in their inventory. It is more a fail-safe so we don't house redundant data files.

nessie/jobs/sync_canvas_requests_snapshots.py
sandeepmjay: makes sense moving it to a shared s3_utils location.

paulkerschen:
Looks good to me once merge conflict is resolved, @sandeepmjay.

PR #606, opened 2019-11-04
[NS-728] Gradescope table data catalog job.

sandeep-jay:
Gradescope storage and catalog job. Ingestion frequency and sub-level organizations are TBD. This gives a good integration and foundation to build upon.


paulkerschen:
Thanks @sandeepmjay!

PR #609, opened 2019-11-22
BOAC-3002: Stores automated term configs in RDS

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3002


johncrossman:
With all this automation, who needs the humans?  Thanks @lyttam

PR #610, opened 2019-11-22
BOAC-3058, keep Nessie in line with BOA package deps

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3058

A few minor upgrades; the diff is substantial only because I removed carets from package "dependencies" section.  I verified Nessie dashboard – no Vue breakages.


paulkerschen:
Nessie's a follower.

PR #624, opened 2019-12-04
NS-755 Break up more SIS data merging by term too

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-755

How about this?


johncrossman:
How about that.

PR #626, opened 2019-12-05
[QA] NS-755: Increases instance size to m5.large

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-755

master PR: #623


lyttam:
I imagine this will cost extra, but qa workers spin up for only 4 hours (vs 8 hours on dev).

lyttam:
Also maybe @paulkerschen should merge this one so it doesn't interfere with his testing on nessie-qa.

paulkerschen:
Per conversation with @lyttam, we're going to hold off on this for the moment to see how nightly jobs run without an instance upgrade.

paulkerschen:
Turns out we need the bigger guns.

PR #628, opened 2019-12-05
NS-757: Decouples SIS terms from the rest of SIS data

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-757

Terms are now in their own schema, which is created on startup (before initializing the schedule) so that the data is available to other jobs and to the front-end APIs that need them.


paulkerschen:
This is a big move, but I suppose it's worth untangling. Thanks, @lyttam - do we need an added release instruction for the local REDSHIFT_SCHEMA_SIS_TERMS config?

lyttam:
@paulkerschen yes, thanks!

PR #631, opened 2019-12-10
BOAC-3118 Handle alternate email addresses in SIS profile merge

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3118

### DIFF
--- a/nessie/merged/sis_profile.py
+++ b/nessie/merged/sis_profile.py
@@ -239,10 +239,11 @@ def merge_sis_profile_emails(sis_student_api_feed, sis_profile):
     for email in sis_student_api_feed.get('emails', []):
         if email.get('primary'):
             primary_email = email.get('emailAddress')
-            break
         elif email.get('type', {}).get('code') == 'CAMP':
             campus_email = email.get('emailAddress')
-    sis_profile['emailAddress'] = primary_email or campus_email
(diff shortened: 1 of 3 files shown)


nessie/merged/sis_profile.py
```diff
@@ -239,10 +239,11 @@ def merge_sis_profile_emails(sis_student_api_feed, sis_profile):
     for email in sis_student_api_feed.get('emails', []):
         if email.get('primary'):
             primary_email = email.get('emailAddress')
-            break
         elif email.get('type', {}).get('code') == 'CAMP':
             campus_email = email.get('emailAddress')
-    sis_profile['emailAddress'] = primary_email or campus_email
+    sis_profile['emailAddress'] = campus_email or primary_email
+    if primary_email and primary_email != campus_email:
+        sis_profile['emailAddressAlternate'] = primary_email
```
lyttam: @paulkerschen If there is no campus email, `emailAddress` and `emailAddressAlternate` will both contain the primary email - is that ok?

paulkerschen: Good catch, @lyttam - may as well amend to catch that case.

lyttam: Looks good, thanks!

PR #634, opened 2019-12-12
BOAC-3131, Nessie gets copyright 2020

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3131


paulkerschen:
If man is still alive, if Nessie can survive...

PR #635, opened 2019-12-12
BOAC-3109 Include Calnet-derived advisor attributes in student feeds

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3109

### DIFF
--- a/nessie/lib/queries.py
+++ b/nessie/lib/queries.py
@@ -40,6 +40,10 @@ def advisor_schema():
     return app.config['REDSHIFT_SCHEMA_ADVISOR']
 
 
+def advisor_schema_internal():
+    return app.config['REDSHIFT_SCHEMA_ADVISOR_INTERNAL']
+
+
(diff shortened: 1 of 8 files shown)


nessie/lib/queries.py
```diff
@@ -40,6 +40,10 @@ def advisor_schema():
     return app.config['REDSHIFT_SCHEMA_ADVISOR']
 
 
+def advisor_schema_internal():
+    return app.config['REDSHIFT_SCHEMA_ADVISOR_INTERNAL']
```
johncrossman: @paulkerschen - Should `config/default.py` have `REDSHIFT_SCHEMA_ADVISOR_INTERNAL`?

paulkerschen: @johncrossman Yes, it already does - we just weren't querying it Nessie-side till now.

johncrossman:
Thanks @paulkerschen

PR #648, opened 2020-02-04
BOAC-3271: Handles plans with missing programs

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3271
https://jira.ets.berkeley.edu/jira/browse/NS-781

### DIFF
--- a/nessie/jobs/generate_merged_student_feeds.py
+++ b/nessie/jobs/generate_merged_student_feeds.py
@@ -239,7 +239,7 @@ class GenerateMergedStudentFeeds(BackgroundJob):
 
             for plan in sis_profile.get('plans', []):
                 if plan.get('status') == 'Active':
-                    rows['student_majors'].append(encoded_tsv_row([sid, plan['program'], plan['description']]))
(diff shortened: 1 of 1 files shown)


nessie/jobs/generate_merged_student_feeds.py
```diff
@@ -239,7 +239,7 @@ def generate_student_profile_feed(self, feed_elements, rows, advisors):
 
             for plan in sis_profile.get('plans', []):
                 if plan.get('status') == 'Active':
-                    rows['student_majors'].append(encoded_tsv_row([sid, plan['program'], plan['description']]))
+                    rows['student_majors'].append(encoded_tsv_row([sid, plan.get('program', None), plan['description']]))
```
johncrossman: @lyttam – Looks good. To verify, the `'description'` key will always be present? No need for `.get(...)`?

lyttam: @johncrossman It should always be preset - at least that's how it's been working.  I could add a `.get` for extra security.

lyttam: Amended.

PR #649, opened 2020-02-07
[NS-786] Ingests OUA Admissions data into Loch. 

sandeep-jay:
Contains the following
1) OUA sql templates
2) Test data generator using faker
3) Nessie job to create schema.
4) Integrations to migrate from sftp folder to daily s3 OUA location. 

https://jira.ets.berkeley.edu/jira/browse/NS-786


paulkerschen:
Thanks for the fake data script, @sandeepmjay!

PR #650, opened 2020-02-07
[NS-789] OUA Nessie RDS tables and indexes for BOA rendering.

sandeep-jay:
Continuation from NS-786. Creates slate OUA student admits RDS tables and relevant indexes known so far. 

 https://jira.ets.berkeley.edu/jira/browse/NS-789.


paulkerschen:
That sure is a lot of columns.

PR #661, opened 2020-03-10
NS-801 Store CalNet instructor data in the loch

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-801


sandeepmjay:
Look good @paulkerschen. Shipping it

PR #668, opened 2020-03-18
NS-802 Move away from misleading old 'student_academic_status' table

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-802

For the moment we are only moving around Redshift tables, leaving RDS in place so BOA won't notice anything.


sandeepmjay:
Thanks for the corrections @paulkerschen.

PR #671, opened 2020-03-20
NS-810: Adds minors to RDS student schema

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-810


sandeepmjay:
Looks good @lyttam. Thanks.

PR #681, opened 2020-04-13
NS-812 Bump SQLAlchemy

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-812


sandeepmjay:
Ah the SQLAlchemy dance, one step two-step. Thanks, @paulkerschen !

PR #691, opened 2020-05-12
NS-840: Upgrades flake8 to 3.8.0

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-840


johncrossman:
Thank you @lyttam  – Nessie is stylin' some clean code.

PR #699, opened 2020-06-04
NS-845: Adds job queue for Canvas API jobs

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-845

### DIFF
--- a/nessie/lib/metadata.py
+++ b/nessie/lib/metadata.py
@@ -30,6 +30,17 @@ from flask import current_app as app
 from nessie.externals import rds, s3
 
 
+def create_canvas_api_import_status(job_id, term_id, course_id, table_name):
+    sql = f"""INSERT INTO {_rds_schema()}.canvas_api_import_job_status
+               (job_id, term_id, course_id, table_name, status, instance_id, created_at, updated_at)
+               VALUES (%s, %s, %s, %s, 'created', %s, current_timestamp, current_timestamp)
+               """
+    return rds.execute(
+        sql,
+        params=(job_id, term_id, str(course_id), table_name, _instance_id()),
+    )
+
+
 def create_canvas_sync_status(job_id, filename, canvas_table, source_url):
     sql = f"""INSERT INTO {_rds_schema()}.canvas_sync_job_status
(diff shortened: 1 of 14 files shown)


johncrossman:
Looks good to me. @sandeepmjay - You want to review?

nessie/lib/metadata.py
```diff
@@ -30,6 +30,17 @@
 from nessie.externals import rds, s3
 
 
+def create_canvas_api_import_status(job_id, term_id, course_id, table_name):
+    sql = f"""INSERT INTO {_rds_schema()}.canvas_api_import_job_status
+               (job_id, term_id, course_id, table_name, status, instance_id, created_at, updated_at)
```
sandeepmjay, line 35: @lyttam Does It make sense to include a column on what type of Canvas API we are importing from ?

sandeepmjay:
Looks good @lyttam. I had posted my thoughts on a minor addition to the import status table. I can merge once you have taken a look. Thanks!

tests/conftest.py
```diff
@@ -84,6 +84,18 @@ def metadata_db(app):
         created_at TIMESTAMP NOT NULL,
         updated_at TIMESTAMP NOT NULL
     )""")
+    rds.execute(f"""CREATE TABLE IF NOT EXISTS {rds_schema}.canvas_api_import_job_status
+    (
+        job_id VARCHAR NOT NULL,
+        course_id VARCHAR NOT NULL,
+        term_id VARCHAR NOT NULL,
+        table_name VARCHAR NOT NULL,
```
sandeepmjay, line 92: Same question as above, including the API we are interacting with, might make this status table general purpose for any of the future API import requirements as well?

nessie/lib/metadata.py
lyttam: @sandeepmjay the `table_name` column will contain values such as 'gradebook_history' and 'grade_change_log' (referencing the tables in the canvas_api_ext_dev schema, but also identifying which API was used).

sandeepmjay: sounds good. Thanks, @lyttam.

PR #700, opened 2020-06-04
NS-845: Replaces Canvas API token with list of tokens

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-845


sandeepmjay:
LGTM @lyttam.

PR #705, opened 2020-06-12
BOAC-3576: Adds author CalNet data to Data Science notes

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3576


paulkerschen:
Looks good!

PR #710, opened 2020-06-16
NS-868: Removes length limit from student level column

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-868


paulkerschen:
Thanks much, @lyttam. Seems like a warning sign that the limit was sometimes 2 characters and sometimes 3 - we're better off without.

PR #716, opened 2020-06-25
NS-873: Plugs the active-inactive enrollment gap

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-873


paulkerschen:
Looks right. Thanks @lyttam!

PR #717, opened 2020-06-26
NS-862: Expands term ID math to account for past millennia

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-862


paulkerschen:
Sands of time.

PR #718, opened 2020-06-29
NS-876: Accomodates degrees awarded in legacy winter terms

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-876


paulkerschen:
March madness.

PR #721, opened 2020-07-06
NS-833: Leverages temp files for storing student profile data sets

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-833


paulkerschen:
ExitStack!

PR #728, opened 2020-07-21
NS-891: Uses temp files to store gradebook API responses

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-891


paulkerschen:
Feed that file.

PR #738, opened 2020-07-29
NS-907: Adds job to transform Canvas API data into multiline JSON

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-907


johncrossman:
Looks good, @lyttam  

I'll give @sandeepmjay a chance to review, if he's available.

sandeepmjay:
Looks good @lyttam. Thanks for your work.  One thing to see would be on the performance side while working with individual s3 files. Let review that once we have a successful run.

PR #747, opened 2020-08-26
NS-941, after qa branch cut: update package/py deps

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-941


johncrossman:
After this is merged I'll commit the `semi: [2, 'never']` eslint rule.

PR #748, opened 2020-08-26
NS-941, eslint 'no semicolons' rule (Vue files, javascript/typescript)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-941

A bit Pythonic, you might say.


paulkerschen:
I feel so much less winked at.

PR #752, opened 2020-09-03
NS-932: Populates external schema with EDL SIS data

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-932

Figured I would just create all of the tables (not just degree progress) since it will make UAT easier.


paulkerschen:
Familiar data fields, yet so different. Let’s give it a go!

PR #760, opened 2020-09-17
NS-960: Limits length of SIDs in accordance with SIS API constraints

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-960

Also reorganized column aliases to reduce confusion.


paulkerschen:
Thanks for cleaning up the aliases, @lyttam.

PR #776, opened 2020-11-05
NS-993: Enforces naming convention for wayward SIS note attachments

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-993


sandeepmjay:
Look good @lyttam. Thanks for the fix!

PR #789, opened 2020-12-08
NS-1032, CreateOUASchema will not raise error if 'no OUA files found'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1032

### DIFF
--- a/nessie/jobs/create_oua_schema.py
+++ b/nessie/jobs/create_oua_schema.py
@@ -38,19 +38,7 @@ class CreateOUASchema(BackgroundJob):
     def run(self):
         app.logger.info('Starting OUA Slate schema creation job...')
         app.logger.info('Executing SQL...')
-        self.migrate_oua_sftp_data()
-        external_schema = app.config['REDSHIFT_SCHEMA_OUA']
-        redshift.drop_external_schema(external_schema)
-        resolved_ddl = resolve_sql_template('create_oua_schema_template.sql')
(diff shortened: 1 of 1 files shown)


nessie/jobs/create_oua_schema.py
```diff
@@ -64,17 +52,31 @@ def migrate_oua_sftp_data(self):
...
+                self.create_rds_tables_and_indexes()
+                app.logger.info('OUA Slate RDS indexes created.')
+                return 'OUA schema creation job completed.'
+
+            else:
+                raise BackgroundJobError('OUA Slate schema creation job failed.')
+
         else:
-            raise BackgroundJobError('No OUA files found in SFTP location today. Skipping OUA data refresh')
+            app.logger.warn('No OUA files found in SFTP location today. Skipping OUA data refresh')
```
paulkerschen, line 68: @johncrossman Looks good, but I might have it `return` this message rather than `app.logger.warn`. If we return a string value, then the message should show up on the Nessie dashboard next to the green status, which is handy for checking whether any files were found.

johncrossman: Thanks @paulkerschen  I'll amend.

paulkerschen:
Thanks @johncrossman! Since we’re assembling a Nessie patch release to deal with job failures, I’d say might as well put this to the QA branch too.

PR #794, opened 2020-12-18
NS-1029, ImportDegreeProgress: Use temp external schema rather than a COPY command

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1029

Sadly, I had to disable unit test because the new SQL syntax is incompatible with Postgres (mock Redshift when pytest).


lyttam:
SAD!

PR #795, opened 2020-12-22
NS-996: adds EPN grade option flag to enrollment term index

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-996


sandeepmjay:
Lppks good @lyttam

PR #808, opened 2021-02-05
NS-1053: Builds json feeds, RDS indexes with acad plan data

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-1053

### DIFF
--- a/nessie/sql_templates/create_edl_schema.template.sql
+++ b/nessie/sql_templates/create_edl_schema.template.sql
@@ -46,6 +46,28 @@ ALTER DEFAULT PRIVILEGES IN SCHEMA {redshift_schema_edl} GRANT SELECT ON TABLES 
 -- Internal tables
 --------------------------------------------------------------------
 
+CREATE TABLE {redshift_schema_edl}.student_academic_plan_index
+SORTKEY (sid)
+AS (
+    SELECT
+      student_id AS sid,
+      academic_program_nm AS program,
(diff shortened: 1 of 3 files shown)


nessie/sql_templates/create_edl_schema.template.sql
```diff
@@ -46,6 +46,28 @@ ALTER DEFAULT PRIVILEGES IN SCHEMA {redshift_schema_edl} GRANT SELECT ON TABLES
...
+      academic_program_nm AS program,
+      academic_plan_nm AS plan,
+      academic_plan_type_cd AS plan_type,
+      academic_subplan_nm AS subplan,
+      academic_program_status_desc AS status
+    FROM {redshift_schema_edl_external}.student_academic_plan_data
+    WHERE academic_plan_type_cd IN ('MAJ', 'SS', 'SP', 'HS', 'CRT', 'MIN')
+);
+
+CREATE TABLE {redshift_schema_edl}.student_academic_plans
```
lyttam, line 63: This table may just be a temporary holding place for the partial profiles - maybe we'll rename it `student_profiles` as we add more data.  Or, it could be a staging area and the JSON feeds contained within will later be incorporated into the larger JSON profiles.

pkerschen: Yes, makes sense to leave this open for now.

PR #810, opened 2021-02-10
NS-1076, ImportRegistrationsHistEnr queries edl_ext.student_registration_term_data if feature_flag

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1076

EDL query results are shoe-horned into existing design pattern: put tsv in S3 prior to populating `hist_enr_term_gpas` and `hist_enr_last_registrations`. There are a handful of TODOs.

PR #811, opened 2021-02-10
NS-1076, show EDL-feature-flag setting on /status page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1076

Follow up to PR #810


pkerschen:
🎏

PR #818, opened 2021-02-12
NS-1045, chunked SQL query to get 300K+ student registrations

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1045

Having `...WHERE sid in (300K+ SIDS)...` is killing the app. That no good. Review with `?w=1`


pkerschen:
download

PR #819, opened 2021-02-12
NS-1078, clean up front-end code: follow basic patterns used by BOA, Diablo

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1078

Less code and, at same time, more maintainable. Freshen the UI.


pkerschen:
Carousel!

PR #821, opened 2021-02-12
NS-1045, import_registrations_hist_enr: Respect batch mode when querying EDL

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1045

We hit another memory issue in nessie-dev. Having EDL respect batch-mode is fine, the transaction is fast. This PR includes column name fixes and minor front-end tweak. (Job ran successfully locally.)


pkerschen:
Respectfully yours, EDL.

PR #826, opened 2021-02-17
NS-1081, get_s3_sis_api_daily_path has opt-in to use 'edl' S3 bucket

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1081

Not all calls to `get_s3_sis_api_daily_path()` should use EDL bucket. Opt-in is safer.

### DIFF
--- a/nessie/sql_templates/create_edl_schema.template.sql
+++ b/nessie/sql_templates/create_edl_schema.template.sql
@@ -51,7 +51,7 @@ SORTKEY (section_id)
 AS (
   WITH edl_classes AS (
     SELECT class_number, semester_year_term_cd, instructional_format_nm, class_section_cd, course_nm, course_title_long_nm, graded_section_flg, instruction_mode_cd,
-      /* TODO We find the max units actually enrolled as a placeholder for SIS max allowable units; this probably isn't 100% conformant. */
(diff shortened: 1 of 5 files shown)


nessie/sql_templates/create_edl_schema.template.sql
```diff
@@ -51,7 +51,7 @@ SORTKEY (section_id)
 AS (
   WITH edl_classes AS (
     SELECT class_number, semester_year_term_cd, instructional_format_nm, class_section_cd, course_nm, course_title_long_nm, graded_section_flg, instruction_mode_cd,
-      /* TODO We find the max units actually enrolled as a placeholder for SIS max allowable units; this probably isn't 100% conformant. */
+      -- TODO: We find the max units actually enrolled as a placeholder for SIS max allowable units; this probably isn't 100% conformant.
```
johncrossman, line 54: Nessie-dev complains:
```
[2021-02-17 21:59:54,304] - WARNING: unterminated /* comment at or near "/* TODO We find the max units actually enrolled as a placeholder for SIS max allowable units"
LINE 12:       /* TODO We find the max units actually enrolled as a p...
               ^
```

I would expect those inline comments to work. cc: @pkerschen

pkerschen: Oh, curious. Let's try the inline syntax then.

PR #827, opened 2021-02-17
NS-1080, fix column names on student_enrollment_data query

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1080

And, I think it was the semi-colon in the inline comment that caused a problem. I tested this SQL file on nessie-dev (hacked it with `vi`) and it works.

### DIFF
--- a/nessie/sql_templates/create_edl_schema.template.sql
+++ b/nessie/sql_templates/create_edl_schema.template.sql
@@ -51,7 +51,7 @@ SORTKEY (section_id)
 AS (
   WITH edl_classes AS (
     SELECT class_number, semester_year_term_cd, instructional_format_nm, class_section_cd, course_nm, course_title_long_nm, graded_section_flg, instruction_mode_cd,
-      -- TODO: We find the max units actually enrolled as a placeholder for SIS max allowable units; this probably isn't 100% conformant.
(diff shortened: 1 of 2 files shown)


nessie/sql_templates/create_edl_schema.template.sql
```diff
@@ -51,7 +51,7 @@ SORTKEY (section_id)
 AS (
   WITH edl_classes AS (
     SELECT class_number, semester_year_term_cd, instructional_format_nm, class_section_cd, course_nm, course_title_long_nm, graded_section_flg, instruction_mode_cd,
-      -- TODO: We find the max units actually enrolled as a placeholder for SIS max allowable units; this probably isn't 100% conformant.
+      -- TODO: We find the max units actually enrolled as a placeholder for SIS max allowable units (probably not entirely conformant)
```
johncrossman, line 54: https://www.postgresql.org/docs/8.0/sql-syntax.html says, "The semicolon (;) terminates an SQL command. It cannot appear anywhere within a command, except within a string constant or quoted identifier."  Apparently, semicolon in the comment terminates the line and the stuff after it is parsed as SQL.

pkerschen:
Semicolons! Good Lord. Thanks for the fixes @johncrossman.

PR #836, opened 2021-02-20
NS-1087, confirmation in UI when user kicks off a Nessie job

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1087

Kick off a job -> random xkcd.

image


pkerschen:
automation_2x

PR #839, opened 2021-02-22
NS-1095, json.dumps to put valid JSON (EDL) in tsv

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1095


pkerschen:
Pump and dump!

PR #842, opened 2021-02-23
NS-1091, email alert when job fails (disabled until we have bCOP creds)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1091

The `BConnected` class, and config scheme, is cloned from Diablo.


pkerschen:
We can all use a little Diablo in us.

PR #843, opened 2021-02-23
NS-1098, jobs that do not finish cleanly are identified (red icon) in UI

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1098

I've also added a 'duration' column.

### DIFF
--- a/nessie/api/admin_controller.py
+++ b/nessie/api/admin_controller.py
@@ -53,19 +53,20 @@ def console_available_jobs():
 @auth_required
 def background_job_status():
     iso_date = request.args.get('date')
-    date_format = '%Y-%m-%d %H:%M'
     date = dateutil.parser.parse(iso_date) if iso_date else datetime.today()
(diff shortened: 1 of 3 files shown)


pkerschen:
@johncrossman Will this actually distinguish between stalled jobs and jobs in progress? I thought the current state of affairs was that when a job starts it gets a “finished” time equal to the “started” time (because it’s just sourcing updated_at from the dB). Which is a bug, but maybe better approached by suppressing finished time equal to start time (and raising the alarm if the job’s been running a couple of hours or more.)

johncrossman:
@pkerschen - Ah, I didn't realize that "when a job starts it gets a finished time equal to the started time." Okay, I'll revise the PR.

johncrossman:
I had to force version on `npx npm-force-resolutions@0.0.3` as shown in https://github.com/rogeriochaves/npm-force-resolutions/issues/22. A temporary workaround, we hope.

PR #847, opened 2021-02-24
NS-1091, put more failed-job info in alert_email

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1091

I also removed redundant `EMAIL_FEATURE_ENABLED` check. The diff is far simpler than it looks: review with `?w=1`.


pkerschen:
The more you know…

PR #849, opened 2021-02-24
NS-1104, API endpoint: /api/analyze_edl/import_registrations/<sid> (nessie-local only)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1104

This controller can be used for various EDL comparisons.


lyttam:
Great idea @johncrossman. FYI, I don't think we'll be able to include `demographics` in the registrations feed for comparison to `api_demographics` because that's the raw API response, and it's an extra step to build that from EDL data as opposed to just building the `demographics` portion of the distilled `student_profile`. But I can work out a way to address that after merging your PR.

PR #855, opened 2021-02-26
NS-1087, very minor cleanup in Nessie UI

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1087


lyttam:
Glad to see this. Now I can read the whole XKCD before it disappears.

PR #856, opened 2021-02-26
NS-1089: Creates empty tables to avoid job failure

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-1089


johncrossman:
Thanks for the correction on `student_schema_table('sis_profiles')`

PR #870, opened 2021-04-08
Adding new 00_ami.config for NS-1143

dkawase:
This is a temporary change that may be reversed in May when the feature is turned on globally. Jira is NS-1143.


johncrossman:
Thanks @dkawase

PR #886, opened 2021-05-21
NS-1151: fetch piazza 'monthly' archive on the 1st day of the month

jblowe:
... and not on the last day of the month.

The timing is alas just a bit complicated!


sandeepmjay:
Looks good @jblowe! Thanks.

PR #910, opened 2021-07-02
NS-1176 Remove misleading uid column from advisor/dept table

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1176

One more follow-up to #908. Traced back to the source, that column turns out to refer to student ids, not advisor ids.


lyttam:
@pauline2k Does the `idx_advisor_departments_uid` need to get removed or changed to `sid`?

pauline2k:
@lyttam Oh yes, it does and now it has been. Thanks for catching.

PR #917, opened 2021-07-13
NS-1209, get_all_student_ids() from EDL if feature-flag

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1209

EDL replacement for `undergrads_schema.students` was the main challenge. The other datasources in play (ASC, COE, and the BOA-sourced `non_current_students`) are still used when EDL feature-flag is true.

### DIFF
--- a/nessie/lib/queries.py
+++ b/nessie/lib/queries.py
@@ -112,10 +112,22 @@ def undergrads_schema():
 
 
 def get_all_student_ids():
-    sql = f"""SELECT sid FROM {asc_schema()}.students
-        UNION SELECT sid FROM {coe_schema()}.students
-        UNION SELECT sid FROM {undergrads_schema()}.students
-        UNION SELECT sid FROM {advisee_schema()}.non_current_students"""
+    if app.config['FEATURE_FLAG_EDL_SIS_VIEWS']:
+        sql = f"""SELECT DISTINCT student_id AS sid
+            FROM {edl_external_schema()}.student_academic_plan_data
+            WHERE
+                academic_career_cd='UGRD'
+                AND academic_program_status_cd='AC'
+                AND academic_plan_type_cd != 'MIN'
+            UNION SELECT sid FROM {asc_schema()}.students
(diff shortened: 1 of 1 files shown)


johncrossman:
Relevant Junction code: https://github.com/ets-berkeley-edu/junction/blob/master/app/models/edo_oracle/bulk.rb#L115

nessie/lib/queries.py
```diff
@@ -112,10 +112,22 @@ def undergrads_schema():
 
 
 def get_all_student_ids():
-    sql = f"""SELECT sid FROM {asc_schema()}.students
-        UNION SELECT sid FROM {coe_schema()}.students
-        UNION SELECT sid FROM {undergrads_schema()}.students
-        UNION SELECT sid FROM {advisee_schema()}.non_current_students"""
+    if feature_flag_edl():
```
pauline2k, line 115: Looks good to me, @johncrossman! If we want to try the granular flagging I introduced earlier, this line would presumably be `if app.config['FEATURE_FLAG_EDL_SIS_VIEWS']"` - not that I'm wedded to that way of doing it if there's a better one.

nessie/lib/queries.py
```diff
@@ -112,10 +112,22 @@ def undergrads_schema():
...
+        sql = f"""SELECT student_id AS sid
```
pauline2k, line 116: @johncrossman Do we need a `SELECT DISTINCT` here? I think the EDL table may sometimes have multiple rows for the same SID even with the restrictions below.

nessie/lib/queries.py
johncrossman: Thanks @pauline2k  Yes, let's go granular. Amendment on the way.

nessie/lib/queries.py
johncrossman: @pauline2k - I assumed UNION would de-dupe the list but I'll add DISTINCT to be safe.

PR #920, opened 2021-07-14
[QA] NS-1200 Fix hardcoded schema references in YCBM template

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1200

QA version of #919.


pauline2k:
flake8 / lint-py commit thrown in for a happy linter.

johncrossman:
Turn that linter frown upside down.

PR #924, opened 2021-07-14
NS-1213, when EDL for unfetched_non_advisees, get basic_attributes from calnet_schema

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1213

Includes a tweak to `get_fetched_non_advisees()` query: `student_academic_plan_data` table has `student_id` column, not `sid`.


johncrossman:
This query ain't right. We don't yet have a non-advisee datasource in the EDL world.

PR #934, opened 2021-07-28
NS-1230: Adds cs_staging schema to CreateEdlSchema job

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-1230


johncrossman:
@lyttam - Do we need a QA PR, too?

PR #981, opened 2021-09-07
NS-1243, create_terms_schema job's datasource varies per EDL feature flag

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1243

The CreateSisTermsSchema job is run at startup and creates a schema dedicated to terms. That pattern stays the same but datasource varies per feature flag.  And, I removed 'SiS_' from naming so we don't wrongly identify it as a legacy schema.

Release instructions:
- rename `REDSHIFT_SCHEMA_SIS_TERMS` and `RDS_SCHEMA_SIS_TERMS` config
- run CreateSisTermsSchema (post deploy) 
- permanently drop `sis_terms_ext_[dev|qa|prod]` (Redshift) and `sis_terms` (RDS)


pauline2k:
Thanks @johncrossman. Rename makes sense, and love that plpythonu.

PR #982, opened 2021-09-08
NS-1243, postgres/python proc lang does not like f-strings

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1243

See Jira for latest stack-trace. Previous version of the postgres function `deduce_term_name()` caused error in Postico console due to f-strings.


pauline2k:
f is for ooF.

PR #985, opened 2021-09-08
NS-1245, EDL term defs: group by term_id, capturing min/max dates

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1245

For example, EDL gives us:
```
2105, 2010-05-24, 2010-07-02
2105, 2010-06-07, 2010-08-13
2105, 2010-06-21, 2010-08-13
2105, 2010-07-06, 2010-08-13
2105, 2010-07-26, 2010-08-13
```
We group by and put one row in RDS:
```
2105, 2010-05-24, 2010-08-13
```


johncrossman:
If we query EDL's `student_academic_terms_data`  then we don't need the fancy group-by.

PR #987, opened 2021-09-08
NS-1240, post_deploy: DROP SCHEMA IF EXISTS sis_terms_ext_[dev|qa|prod]

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1240

Release instructions updated.


nessie/jobs/create_terms_schema.py
```diff
@@ -53,7 +53,7 @@ def create_external_schema(self):
         if not feature_flag_edl:
             # External schema is necessary when pulling term definitions from S3.
             redshift.drop_external_schema(redshift_schema)
-            sql_template = 'create_terms_schema_per_edo_db.template.sql'
+            sql_template = 'create_terms_schema.template.sql'
```
johncrossman, line 56: sql_template above has no equivalent for EDL schema creation, so its name is simplified.

pauline2k: Makes sense.

PR #989, opened 2021-09-09
NS-1301 Tweak flag-dependent behavior in student feed merge

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1301


nessie/jobs/import_registrations.py
```diff
@@ -61,7 +61,7 @@ def run(self, load_mode='new'):
         rows = {
             'term_gpas': [],
             'last_registrations': [],
-            self.demographics_key: [],
+            'api_demographics': [],
```
pauline2k, line 64: This import job always deals with API demographics even if they're not being used in the upstream merge.

PR #1000, opened 2021-09-15
NS-1309, fix instruction_mode ref in edl.student_class_data query

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1309

Good catch, QA dept.


johncrossman:
image

PR #1006, opened 2021-09-16
[QA] NS-1316, unique intended majors please

johncrossman:

master PR is #1005


johncrossman:
I'll fix

PR #1016, opened 2021-09-22
NS-1336: Adds EDL feature flags for advisors and notes

lyttam:
https://jira.ets.berkeley.edu/jira/browse/NS-1336


pauline2k:
Thanks, @lyttam, will merge when green. Release instructions to turn the two new flags on for 5.1?

lyttam:
@pauline2k Yes, thanks. I'll add that.

PR #1023, opened 2021-09-27
NS-1339, ext query to populate edl.academic_standing join on calnet.advisees

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1339

The Junction query is limited to SIDs in s3:calnet-data/sids/snapshot000 so our EDL query must replicate via SQL JOIN.


pauline2k:
Thanks, @johncrossman, makes sense. Will merge once an in-progress job on nessie-dev is done running.

PR #1030, opened 2021-09-28
NS-1344, ed.student_enrollment_data query: include canceled section if student dropped it

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1344

With this modified query, we get the drops reported in Jira above.


pauline2k:
So subtle…

PR #1039, opened 2021-10-05
NS-1346 Save memory with pickles in tempfiles

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1346

Also shuffle some of these job steps into a class hierarchy, because lint-py is rightly complaining that the standalone methods are getting too big to keep straight.


lyttam:
Travis doesn't like pickles.

pauline2k:
Come on, Travis, probiotics!

PR #1052, opened 2021-10-07
NS-1358 Distinguish between career and program status

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1358


johncrossman:
Keep it `simplifier`

PR #1055, opened 2021-10-07
NS-1365 Restrict advisor-student query to existent undergrad advisors

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1365


nessie/sql_templates/edl_create_advisor_schema.template.sql
```diff
@@ -84,7 +84,6 @@ AS (
         academic_career_cd='UGRD'
         AND advisor_id ~ '[0-9]+'
         AND academic_department_cd IS NOT NULL
-        AND academic_plan_cd IS NOT NULL
```
pauline2k, line 87: Null `academic_plan_cd` doesn't occur in the source data.

PR #1066, opened 2021-10-11
NS-1372 Omit unused currentRegistration in merged feeds

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1372


pauline2k:
https://jira.ets.berkeley.edu/jira/browse/NS-1372

PR #1069, opened 2021-10-12
NS-1361, do not str(None) prior to matriculation_term_cd comparison

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1361

...or else you get:
```
  File "/opt/python/current/app/nessie/lib/berkeley.py", line 253, in term_info_for_sis_term_id
    return season_codes[sis_id[3:4]], year
KeyError: 'e'
```


pauline2k:
We put the e in None!

PR #1074, opened 2021-10-18
NS-1368, E&I Advising Notes schema load: a more exact s3 path

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-1368

The job is hitting a Spectrum "retries exceeded" error which, in other scenarios, is due to very large files in S3. In this case, we have 120+ small JSON files. First step is a more exact S3 path, in case there is cost in recursing down from base `e_i` dir.


lyttam:
@johncrossman Perhaps renaming the JSON files with a hashed date would help.

PR #1088, opened 2021-10-25
NS-1371, do not favor UGRD when merge_degrees

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1371


pauline2k:
Let's see what happens!

PR #1095, opened 2021-10-28
NS-1390 Deduce level from future term only as a last resort

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1390

### DIFF
--- a/nessie/lib/queries.py
+++ b/nessie/lib/queries.py
@@ -25,7 +25,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 
 from flask import current_app as app
 from nessie.externals import rds, redshift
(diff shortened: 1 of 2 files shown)


nessie/lib/queries.py
```diff
@@ -419,18 +419,6 @@ def stream_edl_registrations():
...
-                OR
-                (
-                    r.term_enrolled_units IS NOT NULL
-                    AND
-                    (
-                        r.semester_year_term_cd <= '{current_term_id()}'
-                        OR
-                        (r.term_enrolled_units > 0 AND r.semester_year_term_cd > '{current_term_id()}')
-                    )
-                )
```
pauline2k, line 433: With logic added to the Python job, this SQL restriction should be no longer needed.

PR #1097, opened 2021-10-28
[dev] NS-1400, fix 01_download_local_configs.sh (copy/paste strikes again)

johncrossman:

highlands | lowlands


pauline2k:
highlands-lowlands

PR #1098, opened 2021-10-28
NS-1400, nessie-al2 in s3 path to pull configs

johncrossman:

### DIFF
--- a/scripts/deploy_nessie_config.sh
+++ b/scripts/deploy_nessie_config.sh
@@ -35,8 +35,8 @@ else
 fi
 
 # Download from Amazon S3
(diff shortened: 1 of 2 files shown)


scripts/deploy_nessie_config.sh
```diff
@@ -36,7 +36,7 @@ fi
 
 # Download from Amazon S3
 config_name=$(echo "${eb_env}" | sed -E -e 's/(-worker|-master)//')
```
pauline2k: @johncrossman Replace `(-worker|-master)` here too?

johncrossman:
Amended. After discussion with Ops, we agreed to have configs at `s3://la-deploy-configs/nessie-al2-${env_name}.py`

PR #1106, opened 2021-11-04
NS-1232, chown webapp (not wsgi) in scripts/deploy_nessie_config.sh

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1232


pauline2k:
Oh yes, that one.

PR #1122, opened 2021-11-09
NOJIRA, change label of Nessie 'Status' tab

johncrossman:
Previous PR did wake up nessie-dev CodePipeline, as we hoped. This final test is verifying that ets-readonly demotion (w.r.t Github rights) does not break our setup.


johncrossman:
I'll merge this myself, after current CodePipeline is done with nessie-dev deploy.

PR #1133, opened 2021-11-16
BOAC-4540, expand primary key on {rds_schema_student}.student_degrees: sid, plan, term_id

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4540

Follow up to PR #1132

### DIFF
--- a/nessie/sql_templates/create_rds_indexes.template.sql
+++ b/nessie/sql_templates/create_rds_indexes.template.sql
@@ -170,7 +170,7 @@ CREATE TABLE IF NOT EXISTS {rds_schema_student}.student_degrees
     date_awarded VARCHAR NOT NULL,
     term_id VARCHAR(4) NOT NULL,
     hist_enr BOOLEAN NOT NULL,
-    PRIMARY KEY (sid, plan)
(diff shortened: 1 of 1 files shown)


nessie/sql_templates/create_rds_indexes.template.sql
```diff
@@ -170,7 +170,7 @@ CREATE TABLE IF NOT EXISTS {rds_schema_student}.student_degrees
     date_awarded VARCHAR NOT NULL,
     term_id VARCHAR(4) NOT NULL,
     hist_enr BOOLEAN NOT NULL,
-    PRIMARY KEY (sid, plan)
+    PRIMARY KEY (sid, plan, term_id)
```
johncrossman, line 173: `{rds_schema_student}.student_profiles` has seven SIDs where the same plan is mapped to two different terms. For example:
```
Mechanical Engineering   2020-05-15   2202   FALSE
Mechanical Engineering   2021-05-14   2212   FALSE
```

pauline2k: Every possible permutation is in the lake somewhere.

PR #1135, opened 2021-11-17
NS-1421 Remove a whole mess of unused code and Redshift tables

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1421


johncrossman:
image

PR #1156, opened 2021-12-07
NS-1437 Remove legacy SIS schema

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1437


johncrossman:
Too-da-loo SIS schema. We had a good run.

PR #1162, opened 2021-12-08
NS-1437 Cast drop_date as a date indeed

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1437

One more follow-up to #1156.


johncrossman:
It's a date.

PR #1171, opened 2021-12-13
BOAC-2216: Fills profile summary col with initial value, removes default value

lyttam:

### DIFF
--- a/scripts/migrate/20211209-BOAC-2216/redshift_pre_deploy_01_add_column_student_profile_summary.sql
+++ b/scripts/migrate/20211209-BOAC-2216/redshift_pre_deploy_01_add_column_student_profile_summary.sql
@@ -1,6 +1,29 @@
 BEGIN TRANSACTION;
 
-ALTER TABLE student.student_profiles ADD COLUMN profile_summary VARCHAR(max) NOT NULL DEFAULT '{}';
-ALTER TABLE student_staging.student_profiles ADD COLUMN profile_summary VARCHAR(max) NOT NULL DEFAULT '{}';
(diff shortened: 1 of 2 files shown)


scripts/migrate/20211209-BOAC-2216/redshift_pre_deploy_01_add_column_student_profile_summary.sql
```diff
@@ -1,6 +1,29 @@
 BEGIN TRANSACTION;
 
-ALTER TABLE student.student_profiles ADD COLUMN profile_summary VARCHAR(max) NOT NULL DEFAULT '{}';
-ALTER TABLE student_staging.student_profiles ADD COLUMN profile_summary VARCHAR(max) NOT NULL DEFAULT '{}';
```
pauline2k, line 4: @lyttam Do these ALTER TABLE statements not work? I thought that Redshift had finally been updated to support ADD COLUMN.

lyttam: @pauline2k Yes, ADD COLUMN works, but I didn't like having to set a DEFAULT value for it. This way the new column will have value '{}' only temporarily.

The thing that still doesn't work in Redshift is ALTER COLUMN, which would allow me to apply NOT NULL to the column after populating it.

pauline2k: @lyttam Got it, makes sense!

PR #1178, opened 2021-12-22
BOAC-2216: Hitches acad standing, term GPA to existing query

lyttam:

This strategy doesn't add any new queries to the stream.


nessie/jobs/create_edl_schema.py
```diff
@@ -52,7 +52,7 @@ class CreateEdlSchema(BackgroundJob):
 
     def run(self):
         app.logger.info('Starting EDL schema creation job...')
-        self.create_schema()
+        # self.create_schema()
```
pauline2k: @lyttam Is this comment-out a leftover from testing?

lyttam: Whoops, yes! Will amend.

pauline2k:
Let's try it out.

PR #1179, opened 2021-12-22
BOAC-2216: adds academic standing and term GPA to profile (take 2)

lyttam:


pauline2k:
Looks promising…

PR #1189, opened 2022-01-06
BOAC-4606, add history_dept.advising_notes_search_index to aggregated fts_index

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4606

Notes don't exist until they are searchable.


johncrossman:
cc: @pauline2k @lyttam

PR #1190, opened 2022-01-07
BOAC-2216: Fixes bug resulting in null termGpa for everyone

lyttam:

### DIFF
--- a/nessie/jobs/create_edl_schema.py
+++ b/nessie/jobs/create_edl_schema.py
@@ -414,7 +414,8 @@ class ProfileFeedBuilder(ConcurrentFeedBuilder):
                 'status': latest_career_row['acad_standing_status'],
                 'termName': term_name_for_sis_id(latest_career_row['term_id']),
             }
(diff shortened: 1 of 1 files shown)


nessie/jobs/create_edl_schema.py
```diff
@@ -414,7 +414,8 @@ def _term_gpa(row):
                 'status': latest_career_row['acad_standing_status'],
                 'termName': term_name_for_sis_id(latest_career_row['term_id']),
             }
-        feed['termGpa'] = term_gpas.reverse()
+        term_gpas = term_gpas.reverse()
```
pauline2k: @lyttam Is the bug that the Python `reverse` method doesn't return a value? If so, then should this line just read `term_gpas.reverse()` ? Seems like if we keep the assignment we'd still end up with a null value.

lyttam: @pauline2k you're right! Must be quittin' time. I will amend.

PR #1193, opened 2022-01-08
NS-1446 Restore Nessie's external SIS schema

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1446


johncrossman:
Thank you for keeping those old gears spinning.

image

PR #1198, opened 2022-01-11
BOAC-4612, unreliable JOIN clause w.r.t academic_standing_effective_dt_seq

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4612

This rolls back one line of a [recent change in create_edl_schema.template.sql](https://github.com/ets-berkeley-edu/nessie/commit/ff099d5ad03645ff964270df69755c3efe3cde01#diff-1025cd11792b5fdda5d8a7c5ff99fdddbccc0e0f3d3bb74bae835a92326fbf39L103).  

@lyttam - Seem okay? I have not rolled back your change to `(SELECT ...) AS latest_actions` clause.


lyttam:
@johncrossman I think I added that line because there are cases where a student's academic standing is updated more than once in a single day in SIS, but we really only care about the most recent update.

PR #1208, opened 2022-01-14
NS-1448 Remove undergrads schema

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1448


johncrossman:
Zero lines added. 200+ removed. Beautiful.

PR #1226, opened 2022-03-15
BOAC-4650: Fixing linting error

cesarvh:


cesarvh:
Both this and the previous commit belong to https://jira-secure.berkeley.edu/browse/BOAC-4650

I forgot to change my remotes after forking nessie/etc

PR #1231, opened 2022-03-28
BOAC-3952: Added academic division to student majors table

cesarvh:
https://jira-secure.berkeley.edu/browse/BOAC-3952


pauline2k:
Looks good, @cesarvh! One question - the `academic_division_cd` being selected from Redshift probably isn't the value we'll end up displaying in the BOA UI. Were you thinking of adding a translation layer for division names into BOA code? The alternative would be to select the full name in `academic_division_nm` right off the bat.

cesarvh:
<s>@pauline2k Yep, I added that translation in cohort_utils.py</s>

<s>I can also change the nessie code if that would be better, though</s>

Added the change

pauline2k:
Thanks @cesarvh!

PR #1233, opened 2022-03-31
BOAC-3952: Adds redshift migration script + populates new division column

cesarvh:


cesarvh:
https://jira-secure.berkeley.edu/browse/BOAC-3952

PR #1239, opened 2022-05-16
NS-1476 Apparently array literals need to be super-duper escaped when going through unload

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1476

Follow-up to #1238.


johncrossman:
{{{{Looks good}}}}

PR #1252, opened 2022-05-26
BOAC-4695 Preserve decimal precision in index columns on RDS enrollment terms

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4695


johncrossman:
Properly preserve precision.

PR #1257, opened 2022-06-10
NS-1482 Dependency updates

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1482


pauline2k:
Downgrading to Node 14 does get us around the `npm install` bug, but it then introduces new crashes in the front-end lint and build tasks.

PR #1258, opened 2022-06-13
NS-1482 Adjust package.json resolutions format

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1482


pauline2k:
I'm nervous even to advertise it, but the front end is finally building on Travis again.

johncrossman:
Let's give it a go.

PR #1264, opened 2022-08-01
BOAC-4721 Prefer grad to undergrad careers if more recent

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4721

It's difficult to reason out all the possible consequences of this change, so let's just make it on dev and see what comes out the other end.

`w=1` should clear out the noise from fixture file edits.

### DIFF
--- a/nessie/merged/sis_profile.py
+++ b/nessie/merged/sis_profile.py
@@ -78,17 +78,7 @@ def merge_holds(sis_profile_feed, sis_profile):
 
 
 def merge_sis_profile_academic_status(sis_profile_feed, sis_profile):
-    # It is possible to receive multiple academic statuses. We'll prefer an undergraduate enrollment, but
-    # otherwise select the first well-formed status that is not a Law enrollment.
-    academic_status = None
-    for status in sis_profile_feed.get('academicStatuses', []):
-        status_code = status.get('studentCareer', {}).get('academicCareer', {}).get('code')
-        if status_code and status_code == 'UGRD':
-            academic_status = status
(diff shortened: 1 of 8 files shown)


nessie/merged/sis_profile.py
```diff
@@ -78,17 +78,7 @@ def merge_holds(sis_profile_feed, sis_profile):
...
-    # otherwise select the first well-formed status that is not a Law enrollment.
-    academic_status = None
-    for status in sis_profile_feed.get('academicStatuses', []):
-        status_code = status.get('studentCareer', {}).get('academicCareer', {}).get('code')
-        if status_code and status_code == 'UGRD':
-            academic_status = status
-            break
-        elif status_code in {'UCBX', 'GRAD'}:
-            academic_status = status
-            next
```
pauline2k, line 91: In the EDL world we filter on academic careers before constructing the feed, so we'll never get an `academicStatuses` field with multiple statuses.

tox.ini
```diff
@@ -27,7 +27,7 @@ commands = pytest --durations=10 {posargs: -p no:warnings tests}
 # Bottom of file has Flake8 settings
 commands = flake8 {posargs:application.py config consoler.py nessie scripts tests}
 deps =
-    flake8>=4.0.1
+    flake8==4.0.1
```
pauline2k, line 30: Same issue as in https://github.com/ets-berkeley-edu/damien/pull/338

lyttam:
Here goes!

PR #1271, opened 2022-09-20
BOAC-4713, UNION query must have the same number of columns

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4713

image


pauline2k:
dr.sis and dr.grade have a lot to answer for.

PR #1274, opened 2022-09-26
BOAC-4755, avoid empty WHERE clause in student_schema > refresh_from_staging

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4755


pauline2k:
It's always something...

PR #1284, opened 2022-11-01
BOAC-4777, API endpoint to run RefreshBoacCache

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4777

This job typically runs as part of a "chain" but I see no harm in a run-as-standalone-job option.


johncrossman:
@pauline2k or @lyttam – Thanks (in advance) for reviewing.

PR #1288, opened 2022-11-08
BOAC-4785 Remove per-SID failsafe from enrollment term refresh

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4785


johncrossman:
Thank you @pauline2k!

PR #1324, opened 2023-04-14
BOAC-4870 Prefer active to inactive status when both associated with same plan

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4870


johncrossman:
Life is what happens to you while you're busy [coding] plans.

PR #1325, opened 2023-04-18
BOAC-4886 Restore dropped courses to term feeds

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4886

There's a lot to love about Python, but I think this blessed little significant indent mistake (from #1318) is why dropped courses have disappeared in production BOA.


johncrossman:
Bad indentation can sink a thousand ships.

PR #1352, opened 2023-08-22
RIP-338 Patch upstream bug in SISEDO section data

pauline2k:
https://jira-secure.berkeley.edu/browse/RIP-338


johncrossman:
The data, as it should be.

PR #1353, opened 2023-08-24
RIP-312 Add distinct dept_name to EDO-sourced section data

pauline2k:
https://jira-secure.berkeley.edu/browse/RIP-312

Required for Ripley to provision course sites in the right subaccount.

### DIFF
--- a/nessie/sql_templates/update_rds_indexes_sisedo_full.template.sql
+++ b/nessie/sql_templates/update_rds_indexes_sisedo_full.template.sql
@@ -112,6 +112,7 @@ CREATE TABLE IF NOT EXISTS {rds_schema_sis_internal}.edo_sections
     sis_term_id VARCHAR,
     sis_section_id VARCHAR,
     is_primary BOOLEAN,
+    dept_name VARCHAR,
     sis_course_name VARCHAR,
     sis_course_title VARCHAR,
     sis_instruction_format VARCHAR,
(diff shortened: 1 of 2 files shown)


nessie/sql_templates/update_rds_indexes_sisedo_full.template.sql
```diff
@@ -185,11 +187,12 @@ AND edo.cs_course_id = 'cms-version-independent-id';
 -- Copy over past-term sections from EDL-sourced sis_sections table.
 
 INSERT INTO {rds_schema_sis_internal}.edo_sections
-(sis_term_id, sis_section_id, is_primary, sis_course_name, sis_course_title, sis_instruction_format,
+(sis_term_id, sis_section_id, is_primary, dept_name, sis_course_name, sis_course_title, sis_instruction_format,
       sis_section_num, cs_course_id, session_code, instruction_mode, primary_associated_section_id,
       instructor_uid, instructor_name, instructor_role_code, meeting_location,
       meeting_days, meeting_start_time, meeting_end_time, meeting_start_date, meeting_end_date)
-SELECT sis_term_id, sis_section_id, is_primary, sis_course_name, sis_course_title, sis_instruction_format,
+SELECT sis_term_id, sis_section_id, is_primary, split_part(sis_course_name, ' ', 1)  as dept_name,
```
johncrossman, line 194: @pauline2k - would the `split_part` logic work on, for example, sis_course_name = "NUC ENG 124"?

pauline2k: @johncrossman sis_course_name in this source table is represented as 'NUCENG', so the split will work. Over in edo_sections (which we're using for current sections) the department is 'NUC ENG', matching bCourses subaccounts. So there will be an inconsistency between present-term and past-term data, but since we never provision course sites out of past-term data it shouldn't affect workflow.

johncrossman:
Who said data pipelines are complicated?! Pshaw!

image

PR #1355, opened 2023-08-28
BOAC-4726 IF NOT EXISTS is somehow a Redshift syntax error in this context 🤷🏻‍♀️

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4726


pauline2k:
Travis is somehow not running, but this is not the sort of change that Travis does anything with anyhow.

PR #1357, opened 2023-08-29
BOAC-4726 Shorten feed descriptions for a less verbose UI

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4726


johncrossman:
Less is more.

PR #1362, opened 2023-09-19
BOAC-4985: stashes subplan within plan to capture its status

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4985

### DIFF
--- a/nessie/merged/sis_profile.py
+++ b/nessie/merged/sis_profile.py
@@ -263,6 +263,7 @@ def merge_sis_profile_phones(sis_profile_feed, sis_profile):
 def merge_sis_profile_plans(academic_status, sis_profile):
     plans = []
     plans_minor = []
+    # TODO: Remove this line after BOAC-4985 lands in production
     subplans = set()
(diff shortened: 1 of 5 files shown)


nessie/merged/sis_profile.py
```diff
@@ -297,6 +298,12 @@ def merge_sis_profile_plans(academic_status, sis_profile):
             # A plan with no status is considered discontinued. (NS-689)
             plan_feed['status'] = 'Discontinued'
 
+        # Add any subplans.
+        if student_plan.get('academicSubPlan', None):
+            plan_feed['subplan'] = student_plan['academicSubPlan']
```
johncrossman, line 303: @lyttam - We want `subplan`, not `subPlan`?

pauline2k: Camel-cased `academicSubPlan` is a Hub-style feed; lowercase `subplan` is our own feed we're serving up for BOA consumption, so we get to set the conventions.

PR #1387, opened 2024-01-19
BOAC-4984: adds pronouns to student profile feed

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4984


pauline2k (commented):
@lyttam Code looks good but afaik this query won't yet work in nessie-dev Redshift? We tend to point all Nessie tiers to prod EDL for latency reasons, so either we need to temporarily point nessie-dev to EDL QA or wait till 1/25 to land this code change.

lyttam:
@pauline2k oh right, I was going to temporarily point nessie-dev to EDL QA.

PR #1390, opened 2024-01-29
BOAC-3890: adds academic program status to student schema

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-3890

### DIFF
--- a/fixtures/redshift_schema_edl.sql
+++ b/fixtures/redshift_schema_edl.sql
@@ -60,6 +60,16 @@ CREATE TABLE IF NOT EXISTS {redshift_schema_edl}.basic_attributes (
     person_type VARCHAR
 );
 
+CREATE TABLE IF NOT EXISTS {redshift_schema_edl}.student_academic_programs (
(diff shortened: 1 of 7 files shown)


fixtures/redshift_schema_edl.sql
```diff
@@ -1,12 +1,3 @@
-CREATE SCHEMA IF NOT EXISTS {redshift_schema_edl_external};
-
-CREATE TABLE IF NOT EXISTS {redshift_schema_edl_external}.student_academic_plan_data (
-    student_id VARCHAR NOT NULL,
-    academic_career_cd VARCHAR,
-    academic_program_status_cd VARCHAR,
-    academic_plan_type_cd VARCHAR
-);
-
```
pauline2k: @lyttam new tables look good, but we probably don't want to lose this fixture table for test purposes?

lyttam: @pauline2k Ah, it didn't look like it was doing anything but apparently it is! Amended.

PR #1397, opened 2024-04-10
[NS-1574] Secrets Manager integration. 

sandeep-jay:
The function starts a boto3 session using STS and retrieves a secret string from Secrets Manager managed service.


pauline2k:
looks good!

PR #1402, opened 2024-04-26
[NS-1518] Adding Lambda integrations and basic invocation functions.

sandeep-jay:
https://jira-secure.berkeley.edu/browse/NS-1581


johncrossman:
Release the Lambdas!

PR #1418, opened 2024-07-10
[NS-1594] Intermediate Table query migrations using CD2 tables

sandeep-jay:
https://jira-secure.berkeley.edu/browse/NS-1594


johncrossman:
That's some masterful SQL. Well done.

PR #1423, opened 2024-08-06
NS-1608, bumping Nessie to node v20 and python v3.11 [DO NOT MERGE]

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1608

DO NOT MERGE... yet.


package.json
```diff
@@ -10,8 +10,6 @@
     "serve-vue": "vue-cli-service serve --open --host localhost"
   },
   "dependencies": {
-    "ajv": "8.11.0",
-    "ajv-keywords": "5.1.0",
```
johncrossman, line 14: @pauline2k - I don't think the `ajv` dependencies (above) are used. Do you know why they were added in the first place? The purpose of the library is front-end schema/object validation: https://ajv.js.org/guide/getting-started.html

pauline2k: @johncrossman `ajv` and `ajv-keywords` are deeply nested dependencies of other packages, and in the past we've found ourselves having to lock down versions in order to avoid other npm problems. If we can actually do without them in the package.json, that is absolutely all to the good.

johncrossman:
I've struggled with this long enough. I'll come back to it with fresh eyes soon. cc: @felder

PR #1430, opened 2024-08-30
NS-1637 Upgrade to Node 20

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1637


buildspec.yml
```diff
@@ -6,7 +6,7 @@ phases:
       nodejs: 12
       python: 3.8
     commands:
-      - n 16  # Force Node update to v16
+      - n 20  # Force Node update to v16
```
lyttam, line 9: @pauline2k trivial, but this comment should be updated to match.

pauline2k: @lyttam right! PR #1431 consolidates.

PR #1441, opened 2024-10-15
[NS-1648] Fixing the cd2 job names that get displayed on Nessie dashboard

sandeep-jay:


sandeep-jay:
https://jira-secure.berkeley.edu/browse/NS-1648

PR #1473, opened 2025-01-29
BOAC-5790: consolidates students with multiple SIDs in profile index

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5790


pauline2k:
makes sense to me!

PR #1494, opened 2025-03-14
NS-1703 / LA-1090: bCourses Service Dash sql templates, refresh schemas job, job api, default configs

lkvoong:
https://jira-secure.berkeley.edu/browse/NS-1703
https://jira-secure.berkeley.edu/browse/LA-1090


sandeep-jay:
Looks good @lkvoong ! Thank you.

PR #1511, opened 2025-04-07
[QA] NOJIRA: v5.31.1 patch version bump

lyttam:


pauline2k:
@lyttam does this need to be pointed at the qa branch instead of dev?

lyttam:
> @lyttam does this need to be pointed at the qa branch instead of dev?

Oops, yes.

PR #1518, opened 2025-04-23
NS-1727, Ruff for Nessie; no more Flake8

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1727

A lot of inline `noqa` added to avoid code disruption. I fixed low risk violations like use of `format` instead of f-string.


johncrossman:
I'll investigate the failure.

PR #1530, opened 2025-05-15
NS-1733, externals/calendly_api.py and first pass on jobs/import_calendly_api.py

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1733


johncrossman:
@pauline2k - Okay, I removed the fixture decorator and Travis passed. I'll circle back to the fixture mystery when I write the pytest. For now, review at your leisure.

pauline2k:
Thank you for digging in! I wonder if you've uncovered a sneaky bug with how we set up the fixture framework.

PR #1541, opened 2025-05-22
NS-1736, create Calendly external Redshift schema

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1736

As more real data lands in prod Calendly, I will add columns (eg, 'guests' as Struct).


config/default.py
```diff
@@ -306,6 +306,7 @@
 REDSHIFT_SCHEMA_EDL_EXTERNAL = 'External EDL SIS schema name'
 REDSHIFT_SCHEMA_EDL_EXTERNAL_EDW = 'External EDL EDW schema name'
 REDSHIFT_SCHEMA_EDL_EXTERNAL_STAGING = 'External staging EDL SIS schema name'
+REDSHIFT_SCHEMA_EOP_ADVISING_NOTES = 'External EOP Advising Notes schema name'
```
johncrossman, line 309: During this work, I noticed`REDSHIFT_SCHEMA_EOP_ADVISING_NOTES` was missing from `default.py`.

nessie/sql_templates/create_calendly_schema.template.sql
```diff
@@ -39,90 +39,17 @@ CREATE EXTERNAL DATABASE IF NOT EXISTS;
...
+    uri VARCHAR,
+    imported_at VARCHAR
 )
 ROW FORMAT SERDE 'org.openx.data.jsonserde.JsonSerDe'
 WITH SERDEPROPERTIES ('ignore.malformed.json' = 'true')
 LOCATION '{loch_s3_calendly_data_path}/archive';
-
---------------------------------------------------------------------
--- Internal schema
---------------------------------------------------------------------
```
johncrossman, line 58: Internal schema is next.

PR #1565, opened 2025-06-20
NS-1758: add additional advisors to CE3 materialized views

lkvoong:
https://jira-secure.berkeley.edu/browse/NS-1758
https://jira-secure.berkeley.edu/browse/LA-1160


johncrossman:
Looks good to me.

PR #1570, opened 2025-07-02
NS-1777, fix minor regression in import_calendly_api

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1777


nessie/api/status_controller.py
```diff
@@ -34,17 +34,32 @@
...
+    'LOGGING_LEVEL',
+    'LOGGING_PROPAGATION_LEVEL',
     'NESSIE_ENV',
+    'PIAZZA_API_URL',
     'WORKER_QUEUE_ENABLED',
     'WORKER_THREADS',
+    'YCBM_BASE_URL',
+    'YCBM_FETCH_DAYS_AHEAD',
+    'YCBM_FETCH_DAYS_BEHIND',
 ]
```
johncrossman, line 63: ...And, we expose a few more configs in the UI. Useful when smoke-testing a deploy.

PR #1582, opened 2025-08-15
NS-1786, redshift db migrate script to add acad_standing_description to student.academic_standing

johncrossman:
https://jira-secure.berkeley.edu/browse/NS-1786

I'll need @pauline2k to review this one. Afaik, the 'student' schema in Redshift is never dropped and recreated; therefore, my previous schema changes in `create_student_schema.template.sql` (see PR #1581) are never applied because `CREATE TABLE IF NOT EXISTS` does nothing to an existing table.

Is this the proper migrate script given my dilemma? Thanks @pauline2k  :)


pauline2k:
Looks good!

PR #1584, opened 2025-08-26
BOAC-6312 (Nessie), create redshift_schema_edl.academic_standing via term-sensitive CASE stmt

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6312


pauline2k:
Where the elite meet to obsolete.

PR #1624, opened 2026-06-04
[QA] BOAC-5593 Index academic subplans in RDS student schema

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-5593

QA version of #1621.


lyttam:
@pauline2k This one is missing the CLAUDE.md file, is that intentional?

pauline2k:
@lyttam Yes, that was a separate commit and that file's only for local dev work, so there's no urgency about getting it into qa or prod.

